PR #3, opened 2021-08-16
Setup Continuous Integration

### LINKED ISSUE #1: Setup Continuous Ingetration using Github Actions
- [x] Go public
- [x] Replicate Gitlab CI using Github actions on hosted runners
- [x] Delete Gitlab CI

Maybe as new issues: Add CodeCov integration etc.

joergbrech:
... and copy the PR template from the Gitlab repo.

Closes #1


joergbrech:
Ok, so I think I have most of it copied over. Some comments:

 - For some reason, GCC 7.2 crashes with `make -j` when building `EPI_IO`. So for now, there is no parallel compilation for Linux
 - I commented out the Windows build and test *(they were commented out in the Gitlab CI as well, why?)*. To add Windows builds to CI we have to install HDF5
 - I haven't tested yet what the Github pages look like, because that job is restricted to the `main` branch only.

joergbrech:
Ich bin mir noch nicht 100% sicher, ob die C++ coverage reports auch auf Codecov hochgeladen werden: https://app.codecov.io/gh/DLR-SC/memilio/ 

Aktuell taucht da nur die Python coverage auf.

.github/workflows/main.yml
```diff
@@ -0,0 +1,375 @@
...
+  build-cpp:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          echo 'deb http://download.opensuse.org/repositories/science:/dlr/xUbuntu_18.04/ /' | sudo tee /etc/apt/sources.list.d/science:dlr.list
+          wget -O - https://download.opensuse.org/repositories/science:dlr/xUbuntu_18.04/Release.key  | sudo apt-key add -
+          sudo apt-get -qq update
+          sudo apt-get -qq -y install libhdf5-dev libtixi3-dev lcov
```
mknaranja, line 27: I think libtixi can be removed here

.github/workflows/main.yml
```diff
@@ -0,0 +1,375 @@
...
+#      - name: Install ninja
+#        run: choco install -y ninja
+#      - name: Build
+#        run: |
+#          cd cpp
+#          md -Force build-win
+#          cd build-win
+#          cmake --version
+#          cmake -G "Ninja" -DCMAKE_BUILD_TYPE=Release -DEPI_USE_BUNDLED_EIGEN=ON -DEPI_BUILD_EPI_IO=ON -DHDF5_USE_STATIC_LIBRARIES=ON -DHDF5_DIR=${{github.workspace}}\Libraries\HDF5-1.12.0-win64\cmake\hdf5 -Dtixi3_DIR=${{github.workspace}}\Libraries\TIXI-3.1.1-win64\lib\tixi3 ..
+#          cp "${{github.workspace}}\Libraries\TIXI-3.1.1-win64\bin\tixi3.dll" "tests"
```
mknaranja, line 64: remove tixi dependance

.github/workflows/main.yml
```diff
@@ -0,0 +1,375 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Dependencies
+        run: |
+          sudo apt-get -qq update
+          sudo apt-get -qq -y install libhdf5-cpp-100 wget gnupg
+          echo 'deb http://download.opensuse.org/repositories/science:/dlr/xUbuntu_18.04/ /' | sudo tee /etc/apt/sources.list.d/science:dlr.list
+          wget -O - https://download.opensuse.org/repositories/science:dlr/xUbuntu_18.04/Release.key  | sudo apt-key add -
+          sudo apt-get -qq update
+          sudo apt-get -qq -y install libtixi3 g++ python3-pip
```
mknaranja, line 86: -tixi

.github/workflows/main.yml
```diff
@@ -0,0 +1,375 @@
...
+        uses: actions/upload-artifact@v2
+        with:
+          name: test-cpp-coverage-reports
+          path: |
+            cpp/build/tests/report.xml
+            coverage.xml
+            coverage_report
+          retention-days: 1
+
+#  test-cpp-win:
```
mknaranja, line 113: should be reenabled

mknaranja (changes requested):
The windows runner should be reactivated. Please transfer other open problems (i.e., gcc 7.2 issue) to a new issue

joergbrech:
I am a bit at a slump trying to get the Windows builds to work. Currently I am stuck here: https://github.com/DLR-SC/memilio/runs/3350239150?check_suite_focus=true#step:5:242 

Apparently, the copy assignment operator of `epi::RowMajorIterator<Eigen::Array<epi::UncertainValue,-1,1,0,-1,1>,false>` was implicitly deleted because it stores a reference to the array, but it is needed by `std::copy` here:

https://github.com/DLR-SC/memilio/blob/99bfd7a4b4d8393cf1337893acaa9997a7f8853b/cpp/epidemiology/utils/custom_index_array.h#L622

@dabele, @WadimKoslow, do you have any idea how to fix this?

joergbrech:
> 
> 
> I am a bit at a slump trying to get the Windows builds to work. Currently I am stuck here: https://github.com/DLR-SC/memilio/runs/3350239150?check_suite_focus=true#step:5:242
> 
> Apparently, the copy assignment operator of `epi::RowMajorIterator<Eigen::Array<epi::UncertainValue,-1,1,0,-1,1>,false>` was implicitly deleted because it stores a reference to the array, but it is needed by `std::copy` here:
> 
> https://github.com/DLR-SC/memilio/blob/99bfd7a4b4d8393cf1337893acaa9997a7f8853b/cpp/epidemiology/utils/custom_index_array.h#L622
> 
> @dabele, @WadimKoslow, do you have any idea how to fix this?

Nevermind, I think I found the issues. 

Still a linking problem with boost, I will check it out tomorrow

.github/workflows/main.yml
joergbrech: Finally! :goberserk: @mknaranja, could you have another look?

BTW, I also figured out why the C++ coverage was missing from codecov and fixed that.

cpp/epidemiology/secir/contact_matrix.h
```diff
@@ -300,14 +300,12 @@ class DampingMatrixExpressionGroup
         assert(il.size() > 0);
     }
 
-private:
```
mknaranja, line 303: I don't know if we should remove these scopes here. Maybe the issue was that there were functions in public, private, public?

@dabele Could you check that everyting still is as intended?

cpp/epidemiology/utils/eigen_util.h
```diff
@@ -206,9 +206,11 @@ namespace details
 {
     //true if elements returned by matrix(i, j) are references where matrix is of type M;
     //false if the elements are temporaries, e.g. for expressions like Eigen::MatrixXd::Constant(r, c, v);
+    //Caution: This works only for matrix expressions having the LinearAccessBit:
+    // https://eigen.tuxfamily.org/dox/classEigen_1_1DenseCoeffsBase_3_01Derived_00_01ReadOnlyAccessors_01_4.html#a496672306836589fa04a6ab33cb0cf2a
     template <class M>
     using IsElementReference =
-        std::is_reference<decltype(std::declval<M>()(std::declval<Eigen::Index>(), std::declval<Eigen::Index>()))>;
```
mknaranja, line 211: I don't know if this now always works as intended.

@dabele Could you check that everyting still is as intended?

cpp/epidemiology_io/json_serializer.h
```diff
@@ -601,15 +601,15 @@ IOResult<T> read_json(const std::string& path, Tag<T> tag, int flags = IOF_None)
 //Implementations for JsonContext/Object member functions below//
 /////////////////////////////////////////////////////////////////
 
-template <class T, std::enable_if_t<JsonType<T>::value, void*> = nullptr>
```
mknaranja, line 604: I don't know if this is correct SFINAE anymore.

@dabele Could you check that everyting still is as intended?

mknaranja (commented):
There are still some other comments on this PR from your side and where I'm not 100% sure which of these are arleady resolved ;-)

cpp/epidemiology/secir/contact_matrix.h
joergbrech: I don't know why the constructor was declared private. I removed the declaration because for MSVC one of the derived classes could not access the private constructor of its base. I assume that in gcc the public initializer list ctor is used, while in MSVC this implicit conversion does not work.

cpp/epidemiology/utils/eigen_util.h
joergbrech: To be honest, I am not 100% sure why this works with gcc. This check is part of the deserialization of `CustomIndexArray` that stores a flat vector, so MSVC complained that no operator exists accepting two inputs. As long as we only use matrix expressions that can be accessed with the `[]` operator linearly, this check should work for all arrays with contiguous memory.

cpp/epidemiology_io/json_serializer.h
joergbrech: It is. MSVC gives me a "warning treated as error", because the default template argument was doubly defined: 

https://github.com/DLR-SC/memilio/blob/81b9a1494820e1c6d9c2f34aad4f957a4aa66acd/cpp/epidemiology_io/json_serializer.h#L280

It suffices to define it in the definition.

.github/workflows/main.yml
```diff
@@ -35,40 +33,49 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: build-tests-linux
-          path: cpp/build/tests
+          path: cpp/build
```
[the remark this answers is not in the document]
joergbrech: This was the reason, why the code coverage for cpp did not work. We need to artifact the whole build folder for that, not just the tests subdirectory. This is a bit of a pain, because it is quite big, but it is a working solution. I propose to only try to fix this if we are too annoyed by this CI step taking to long. IMHO this should not be a blocker for this PR.

joergbrech:
> 
> 
> There are still some other comments on this PR from your side and where I'm not 100% sure which of these are arleady resolved ;-)

I don't have any more open points.

.github/workflows/main.yml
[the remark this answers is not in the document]
mknaranja: OK, so we keep in mind to track the time needed for the CI from time to time :)

cpp/epidemiology/utils/eigen_util.h
mknaranja: Ok. I'd like to have @dabele checking it after his vacation anyway :)

mknaranja (approved):

cpp/epidemiology/utils/eigen_util.h
dabele: the expression `v(i, j)` should be valid for matrices as well as vectors, I assume to allow this kind of generic programming, see here: https://godbolt.org/z/zqz4qazdq . And I'm pretty sure it's used in other places as well.

the `(i, j)` accessor is more general than the `[i]` accessor, which is why I used that. `RowMajorIterator` is intended to be able to iterate over any matrix or vector expression.

As for why it doesn't work in MSVC, no idea yet.

cpp/epidemiology/utils/eigen_util.h
```diff
@@ -255,6 +257,23 @@ class RowMajorIterator
...
+     * Default copy constructor and copy assignment operator, default move constructor and move assignment operator
+     */
+    RowMajorIterator(const RowMajorIterator& other)
+        : m_matrix(other.m_matrix)
+        , m_i(other.m_i)
+    {
+    }
+    RowMajorIterator& operator=(const RowMajorIterator& other)
+    {
+        m_matrix = other.m_matrix;
```
dabele, line 270: this is a bug. not your change, that just uncovered it. 

this assignment doesn't rebind the reference but overwrites the contents of the referenced matrix. The iterator needs to store a shallow reference, like a pointer or reference_wrapper. then the implicit constructor/operator works.

dabele (commented):
since this is already merged, I'm going to open a new issue to adress the problems (at least the bug, the others may be acceptable)

PR #20, opened 2021-08-26
Add testing of persons to abm

### LINKED ISSUE #6: add testing of persons to ABM
Currently, there is a single parameter that represents the likelihood of an infection being detected. This doesn't take into account different testing schemes, false positives, etc, and the parameter value is difficult to infer from real world data, so probably would have to be fitted. It's better to model real testing.

* add the ability to test a person for the infection with parameters that represent the quality of the test, e.g. sensitivity and specificity. Positive tests lead to quarantine. 
* add testing for infected persons (ie. persons with symptoms) with a specified probability/rate.
* add testing scheme with parameters location, rate/interval, probability (maybe others) to regularly test persons so it's possible to define e.g. testing at work once a week.
* expand the abm example for testing.

elisabeth96:

### Checks by code author:

Closes #6


cpp/epidemiology/abm/person.cpp
```diff
@@ -115,4 +125,29 @@ bool Person::goes_to_school(TimePoint t, const AbmMigrationParameters& params) c
...
+            return false;
+        }
+    }
+    else {
+        if (random < specificity){
+            m_quarantine = false;
+            return false;
+        }
+        else {
+            m_quarantine = true;
```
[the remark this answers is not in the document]
elisabeth96: There are two possibilities: 
- A person with a false positive can leave quarantine if they are tested again and get a negative result. (But at the moment they do not get tested. One could introduce regular tests for people in quarantine after a certain time?)
- If the `Person::interact` function is called everyone that is not infected leaves quarantine automatically. I think this depends on the time steps of the simulation so there might be a better solution?

cpp/epidemiology/abm/world.cpp
```diff
@@ -82,6 +82,7 @@ void World::migration(TimePoint t, TimeSpan dt)
                 auto target_type = rule.first(*person, t, dt, m_migration_parameters);
                 Location* target      = find_location(target_type, *person);
                 if (target != & get_location(*person)) {
+                    target->get_testing_scheme().run_scheme(*person, m_testing_parameters);
```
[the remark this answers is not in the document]
elisabeth96: Yes, that makes sense.

It would be nice to combine the testing scheme with the migration rules, but one needs the location of the target location to run the testing scheme. I think, to get the location one needs the world in migration rules. If I remember correctly, we do not want to do this?

I added a check in `World::migration` such that a positive tested person cannot migrate. But I am open to find a nicer solution.

cpp/epidemiology/abm/person.cpp
```diff
@@ -25,30 +25,37 @@
...
-    m_random_workgroup = UniformDistribution<double>::get_instance()();
+    m_random_workgroup   = UniformDistribution<double>::get_instance()();
     m_random_schoolgroup = UniformDistribution<double>::get_instance()();
-    if (state == InfectionState::Infected_Detected) {
-        m_quarantine = true;
+    if (infection_properties.state == InfectionState::Infected) {
+        if (infection_properties.detected) {
+            m_quarantine = true;
+        }
     }
```
dabele, line 44:
```suggestion
    if (infection_properties.state == InfectionState::Infected && infection_properties.detected) {
        m_quarantine = true;
    }
```

cpp/epidemiology/abm/person.cpp
```diff
@@ -66,9 +73,15 @@ void Person::interact(TimeSpan dt, const GlobalInfectionParameters& global_infec
         }
     }
 
-    if (new_state == InfectionState::Infected_Detected || new_state == InfectionState::Infected_Severe || new_state == InfectionState::Infected_Critical) {
+    if (new_state == InfectionState::Infected_Severe || new_state == InfectionState::Infected_Critical) {
         m_quarantine = true;
     }
+    else if (new_state == InfectionState::Infected) {
+        double rand = UniformDistribution<double>::get_instance()();
+        if (rand < global_infection_params.get<TestWhileInfected>()[this->m_age] / dt.days()) {
```
dabele, line 81: doesn't this need to be `p * t` instead of `p / t` (p is probability, t is time step)? Probability per time step is lower if the time step is shorter.

cpp/epidemiology/abm/person.cpp
```diff
@@ -115,4 +128,31 @@ bool Person::goes_to_school(TimePoint t, const AbmMigrationParameters& params) c
...
+    else {
+        if (random < params.specificity) {
+            m_quarantine               = false;
+            m_time_since_negative_test = days(0);
+            return false;
+        }
+        else {
+            return true;
+        }
+    }
```
dabele, line 156: maybe I'm misremembering, or may you aren't finished with the changes yet, but didn't we agree to ignore tests of healthy persons until we do re-testing to get them out of quarantine?

dabele (changes requested):

cpp/epidemiology/abm/person.cpp
elisabeth96: Oh, I thought we agreed on not putting healthy persons in quarantine (even if they tested positive) but keep testing them. Do you think we should not test them at all/ just give all healthy people a positive test result?

cpp/epidemiology/abm/person.cpp
elisabeth96: Yes, I will change that

cpp/epidemiology/abm/person.cpp
dabele: I think you mean "a negative test result". that was what I was thinking about. but this solution is fine, we get at least part of the effect of false positives.

dabele (approved):

PR #26, opened 2021-09-01
rowmajoriterator assignment overwrites referenced matrix

### LINKED ISSUE #23: RowMajorIterator assignment overwrites referenced matrix
pull request #3 uncovered a bug in `RowMajorIterator`.

the assignment operator of `RowMajorIterator` doesn't rebind the reference but overwrites the contents of the referenced matrix. The iterator needs to store a shallow reference, like a pointer or reference_wrapper. then the implicit constructor/operator works.

For most uses of the iterator, this is not recognizable, since assignee and assigned value reference the same matrix.

Other possible issues that were introduced by the PR and may be adressed at the same time:
- eigen matrix utility IsElementReference uses `[i]` operator instead of `(i,j)` operator.
- DampingMatrixExpression private constructor was made public

_Originally posted by @dabele in https://github.com/DLR-SC/memilio/pull/3#discussion_r698603071_

dabele:
## Changes

- `RowMajorIterator` stores pointer to the matrix that is iterated over instead of a reference so that default copies and assignments of iterators are always shallow as intended
- Revert utility `IsElementReference` to check for operator `(i, j)` instead of operator `[i]`. The multi-index function call operator is more general and works for all matrix expressions. the check works fine in local gcc and msvc builds, CI, and compiler explorer, so I'm not sure what the issue was.

### Checks by code author:

Closes #23

### DIFF
--- a/cpp/epidemiology/utils/eigen_util.h
+++ b/cpp/epidemiology/utils/eigen_util.h
@@ -222,8 +222,9 @@ template <class M, bool IsConst = false>
 class RowMajorIterator
 {
 public:
-    using MatrixReference = std::conditional_t<IsConst, const M&, M&>;
-    static_assert(IsConst || details::IsElementReference<MatrixReference>::value,
+    using MatrixRef = std::conditional_t<IsConst, const M&, M&>;
+    using MatrixPtr = std::conditional_t<IsConst, const M*, M*>;
+    static_assert(IsConst || details::IsElementReference<M>::value,
                   "Iterator must be const if matrix is not in memory.");
 
     using iterator_category = std::random_access_iterator_tag;
@@ -242,7 +243,7 @@ public:
             return &value;
         }
     };
-    using pointer = std::conditional_t<details::IsElementReference<MatrixReference>::value,
+    using pointer = std::conditional_t<details::IsElementReference<MatrixRef>::value,
                                        std::conditional_t<IsConst, const value_type*, value_type*>, Proxy>;
 
     /**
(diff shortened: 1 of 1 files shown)


dabele:
The constructor that was made public in PR #3 was not made private. MSVC handles the declaration `using Base::Base` differently than gcc. I think gcc is correct if I understand the wording on [cppreference](https://en.cppreference.com/w/cpp/language/using_declaration) correctly: constructors are all inherited regardless of their access modifier, but the inherited constructor is only visible where the base class constructor would be visible. e.g. a private constructor that is inherited through a using declaration is only visible from within the base class, not from within the derived class. This is precisely how we used it in the templated deserialize function that is defined in the base class but that can construct an instance of a derived class. I think this feature of the standard is intended to support CRTP type things without having to expose anything. In MSVC, this doesn't work, private or even protected constructors that are inherited by the derived class are not visible in the base class, only in the derived class, the constructor has to be public. 

Since it's not a huge deal to have this public constructor, I left it as it is.

cpp/epidemiology/utils/eigen_util.h
```diff
@@ -242,7 +241,7 @@ class RowMajorIterator
             return &value;
         }
     };
-    using pointer = std::conditional_t<details::IsElementReference<MatrixReference>::value,
+    using pointer = std::conditional_t<details::IsElementReference<MatrixRef>::value,
```
joergbrech, line 244: Not a big issue, I was just wondering: Is `MatrixRef` still needed? As far as I can tell it is only used here and in the definition of the ctor. Here I believe we can use `MatrixPtr` as well, and for the ctor just `M&`?

joergbrech (approved):
Looks good! Sorry for enforcing this update in #3 without consulting you first, but I felt it was important to have a working CI fast so that everyone can be productive.

cpp/epidemiology/utils/eigen_util.h
dabele: MatrixRef does contain const-ness. A const iterator shouldn't require a non-const reference in the constructor and it's possible that const `operator()` returns a value where non-const `operator()` returns a reference.

dabele:
> Looks good! Sorry for enforcing this update in #3 without consulting you first, but I felt it was important to have a working CI fast so that everyone can be productive.

No problem, of course that was the highest priority.

PR #28, opened 2021-09-01
Resolve 24: SEH exception in dynamicnpis unittests

### LINKED ISSUE #24: SEH exception in dynamicnpis unittests
Unittests for dynamic NPIs crash with an Unknown SEH exception, code 0x00000005. I think it happens only in msvc debug builds.

dabele:
## Changes

Looping over all dampings using iterators doesn't work since the damping referenced by the current iterator may be removed in the loop body, which invalidates the iterator according to the standard. GCC has no problems with this since the iterator is not dereferenced after removing, only incremented, which is not a problem in any normal implementation of iterators. But MSVC in debug builds has advanced safety features for iterators that enforce the invalidation of iterators so an iterator pointing to an element that is removed cannot be incremented anymore. This is technically correct according to the c++ standard, but unexpected, since vector iterators usually behave like pointers that can always be incremented or decremented.

So I changed the loop to use indices instead of iterators. Iterators are still used within the loop body, but the iterator is not used after the element is removed.

### Checks by code author:

closes #24


joergbrech (approved):
Thanks for the explanation. Looks good!

PR #100, opened 2021-10-19
71 improve migration rules in ABM

### LINKED ISSUE #71: Improve migration rules in ABM
Currently, persons (agents) in the ABM follow a very simple routine. The python model has more advanced rules. Please analyze what rules are missing and implement them. Some rules may be implemented differently (or even wrong), those need to be discussed and maybe changed. Properties of persons or locations can be added if needed.

### LINKED ISSUE #111: Add the variable going to school/work time
The agents should go to school/work on random times, in this issue the feature should be added to make the time of going to school/work dependent on the person. Therefore a random variable should be added to each person and the time of going to work should be calculated based on that. Consider that this feature should be adopted to the different time steps in case the time step is not 1 hour but smaller or bigger.

majabedi:

### Checks by code author:
* [x] The following questions are addressed in the documentation*:  Developers (what did you do?, how can it be maintained?), For users (how to use your work?), For admins (how to install and configure your work?)

Closes #111

### DIFF
--- a/cpp/models/abm/migration_rules.cpp
+++ b/cpp/models/abm/migration_rules.cpp
@@ -49,8 +49,9 @@ LocationType random_migration(const Person& person, TimePoint t, TimeSpan dt, co
 LocationType go_to_school(const Person& person, TimePoint t, TimeSpan /*dt*/, const AbmMigrationParameters& params)
 {
     auto current_loc = person.get_location_id().type;
+
     if (current_loc == LocationType::Home && t < params.get<LockdownDate>() && t.day_of_week() < 5 &&
-        t.hour_of_day() >= 8 && person.get_age() == AbmAgeGroup::Age5to14 && person.goes_to_school(t, params) &&
+        t.hour_of_day() == person.get_go_to_school_hour(params) && person.get_age() == AbmAgeGroup::Age5to14 && person.goes_to_school(t, params) &&
         !person.is_in_quarantine()) {
         return mio::LocationType::School;
     }
--- a/cpp/models/abm/migration_rules.cpp
+++ b/cpp/models/abm/migration_rules.cpp
@@ -46,12 +46,13 @@ LocationType random_migration(const Person& person, TimePoint t, TimeSpan dt, co
     return current_loc;
 }
 
-LocationType go_to_school(const Person& person, TimePoint t, TimeSpan /*dt*/, const AbmMigrationParameters& params)
+LocationType go_to_school(const Person& person, TimePoint t, TimeSpan dt, const AbmMigrationParameters& params)
 {
     auto current_loc = person.get_location_id().type;
+    int seconds_of_day = 60*60*24;
 
     if (current_loc == LocationType::Home && t < params.get<LockdownDate>() && t.day_of_week() < 5 &&
-        t.hour_of_day() == person.get_go_to_school_hour(params) && person.get_age() == AbmAgeGroup::Age5to14 && person.goes_to_school(t, params) &&
+        (t.seconds() - person.get_go_to_school_time(params, dt).seconds())%(seconds_of_day) == 0 && person.get_age() == AbmAgeGroup::Age5to14 && person.goes_to_school(t, params) &&
         !person.is_in_quarantine()) {
         return mio::LocationType::School;
     }
@@ -62,13 +63,14 @@ LocationType go_to_school(const Person& person, TimePoint t, TimeSpan /*dt*/, co
     return current_loc;
 }
 
-LocationType go_to_work(const Person& person, TimePoint t, TimeSpan /*dt*/, const AbmMigrationParameters& params)
+LocationType go_to_work(const Person& person, TimePoint t, TimeSpan dt, const AbmMigrationParameters& params)
 {
     auto current_loc = person.get_location_id().type;
+    int seconds_of_day = 60*60*24;
 
     if (current_loc == LocationType::Home && t < params.get<LockdownDate>() &&
         (person.get_age() == AbmAgeGroup::Age15to34 || person.get_age() == AbmAgeGroup::Age35to59) &&
-        t.day_of_week() < 5 && t.hour_of_day() == person.get_go_to_work_hour(params) && person.goes_to_work(t, params) && !person.is_in_quarantine()) {
+        t.day_of_week() < 5 && (t.seconds() - person.get_go_to_work_time(params, dt).seconds())%(seconds_of_day) == 0 && person.goes_to_work(t, params) && !person.is_in_quarantine()) {
         return mio::LocationType::Work;
     }
     //return home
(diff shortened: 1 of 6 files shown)


cpp/models/abm/person.cpp
```diff
@@ -137,6 +140,20 @@ bool Person::goes_to_work(TimePoint t, const AbmMigrationParameters& params) con
     return m_random_workgroup < params.get<WorkRatio>().get_matrix_at(t.days())[0];
 }
 
+int Person::get_go_to_work_hour(const AbmMigrationParameters& params) const
+{
+    int Minimum_Goto_Work_time = params.get<GotoWorkTimeMinimum>()[m_age];
+    int Maximum_Goto_Work_time = params.get<GotoWorkTimeMaximum>()[m_age];
+    return Minimum_Goto_Work_time+(int) ((Maximum_Goto_Work_time-Minimum_Goto_Work_time+1)*m_random_goto_work_hour);
```
dabele, line 147: If I understand this correctly, you are adding `1` to `max - min` so the result after integer truncating can be `max`. I think there is a better way to do this. Also, the time step of the simulation may not be fixed anyway, so rounding to whole hours may cause problems.

I would suggest doing the calculation using `TimeSpan`, which has a better resolution than hours and is semantically clearer. so `t_work = tmin + (tmax - tmin) * r`. Then either go to work if `t + 0.5 * dt >= t_work` or if `t >= t_work`, not sure which is better, just a question of which rounding to use. probably `t + 0.5 * dt >= t_work` is better, since the expected value will be `(max + min) * 0.5`.

cpp/models/abm/person.cpp
```diff
@@ -137,6 +140,20 @@ bool Person::goes_to_work(TimePoint t, const AbmMigrationParameters& params) con
     return m_random_workgroup < params.get<WorkRatio>().get_matrix_at(t.days())[0];
 }
 
+int Person::get_go_to_work_hour(const AbmMigrationParameters& params) const
+{
+    int Minimum_Goto_Work_time = params.get<GotoWorkTimeMinimum>()[m_age];
```
dabele, line 145: we use all lower case variable names.

cpp/models/abm/parameters.h
```diff
@@ -224,10 +224,43 @@ struct SocialEventRate {
     }
 };
 
+struct GotoWorkTimeMinimum {
+    using Type = CustomIndexArray<int, AbmAgeGroup>;
```
dabele, line 228: I think this could be `TimeSpan` instead of `int`. Semantically more consistent.

dabele (changes requested):

majabedi:
Closes #111 totally and only one part of #71

cpp/models/abm/parameters.h
majabedi: This is done.

cpp/models/abm/person.cpp
majabedi: done

cpp/models/abm/person.cpp
```diff
@@ -137,6 +140,24 @@ bool Person::goes_to_work(TimePoint t, const AbmMigrationParameters& params) con
     return m_random_workgroup < params.get<WorkRatio>().get_matrix_at(t.days())[0];
 }
 
+TimeSpan Person::get_go_to_work_time(const AbmMigrationParameters& params, TimeSpan dt) const
+{
+    TimeSpan minimum_goto_work_time = params.get<GotoWorkTimeMinimum>()[m_age];
+    TimeSpan maximum_goto_work_time = params.get<GotoWorkTimeMaximum>()[m_age];
+    int timeSlots = (maximum_goto_work_time.seconds() - minimum_goto_work_time.seconds()) / dt.seconds();
+    int seconds_after_minimum = int (timeSlots * m_random_goto_work_hour) * dt.seconds();
+    return minimum_goto_work_time+ mio::seconds(seconds_after_minimum);
```
dabele, line 149: I don't quite understand why you need to use dt here. doesn't it cancel out?

dabele (changes requested):

cpp/models/abm/migration_rules.cpp
```diff
@@ -46,11 +46,13 @@ LocationType random_migration(const Person& person, TimePoint t, TimeSpan dt, co
...
 
-LocationType go_to_school(const Person& person, TimePoint t, TimeSpan /*dt*/, const AbmMigrationParameters& params)
+LocationType go_to_school(const Person& person, TimePoint t, TimeSpan dt, const AbmMigrationParameters& params)
 {
     auto current_loc = person.get_location_id().type;
+    int seconds_of_day = 60*60*24;
+
     if (current_loc == LocationType::Home && t < params.get<LockdownDate>() && t.day_of_week() < 5 &&
-        t.hour_of_day() >= 8 && person.get_age() == AbmAgeGroup::Age5to14 && person.goes_to_school(t, params) &&
+        (t.seconds() - person.get_go_to_school_time(params, dt).seconds())%(seconds_of_day) == 0 && person.get_age() == AbmAgeGroup::Age5to14 && person.goes_to_school(t, params) &&
```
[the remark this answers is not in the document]
majabedi: I agree.
@dabele  Does this method(time_since_midnight) exist or I create it?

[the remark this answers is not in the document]
dabele: does not exist yet but should be easy to implement, similar to the `hour_of_the_day()` method but not rounding to hours and returning a TimeSpan instead of int.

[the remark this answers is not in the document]
majabedi: Ok. This is done.

cpp/models/abm/migration_rules.cpp
```diff
@@ -61,13 +63,14 @@ LocationType go_to_school(const Person& person, TimePoint t, TimeSpan /*dt*/, co
...
-LocationType go_to_work(const Person& person, TimePoint t, TimeSpan /*dt*/, const AbmMigrationParameters& params)
+LocationType go_to_work(const Person& person, TimePoint t, TimeSpan dt, const AbmMigrationParameters& params)
 {
     auto current_loc = person.get_location_id().type;
+    int seconds_of_day = 60*60*24;
 
     if (current_loc == LocationType::Home && t < params.get<LockdownDate>() &&
         (person.get_age() == AbmAgeGroup::Age15to34 || person.get_age() == AbmAgeGroup::Age35to59) &&
-        t.day_of_week() < 5 && t.hour_of_day() >= 8 && person.goes_to_work(t, params) && !person.is_in_quarantine()) {
+        t.day_of_week() < 5 && (t.seconds() - person.get_go_to_work_time(params, dt).seconds())%(seconds_of_day) == 0 && person.goes_to_work(t, params) && !person.is_in_quarantine()) {
```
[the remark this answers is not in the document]
majabedi: Done

cpp/models/abm/person.cpp
majabedi: with the new changes that I did above I also could change this method. So it does not have the dt anymore!

dabele (approved):

PR #116, opened 2021-11-02
77 ensure unit test coverage of secir get_migration_factors

### LINKED ISSUE #77: get_migration_factors not covered in tests
according to codecov, `get_migration_factors` in secir.h is not covered by unit tests at all. Investigate why and fix it.

dabele:
## Changes

- migration unittest uses the right simulation class
- additional unittest specifically for get_migration_factors

closes #77 

### Checks by code author:

### DIFF
--- a/cpp/memilio/math/floating_point.h
+++ b/cpp/memilio/math/floating_point.h
@@ -20,6 +20,10 @@
 #ifndef EPI_MATH_FLOATING_POINT_H
 #define EPI_MATH_FLOATING_POINT_H
 
+#include <algorithm>
+#include <cmath>
+#include <limits>
+
 namespace mio
 {
     /**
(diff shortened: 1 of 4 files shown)


joergbrech (approved):
Out of curiosity, why did you replace the `auto&` with `auto&&`?

dabele:
> Out of curiosity, why did you replace the `auto&` with `auto&&`?

At first because I didn't understand why `auto&` works at all, since `cast` returns a temporary. After you asked me I have found out that it works because `cast` returns `const CastXpr`, so `auto&` becomes `const CastXpr&`, which is fine since const ref extends lifetime of a temporary. That seems a bit brittle to me, the return type could easily change (intentionally or not) to be non-const in future Eigen versions. It could be `auto` without any ref, but that would be brittle too, since `cast` may return a reference to the original object if the type of the cast is identical to the type in the array. So if we ever change the parameters from `UncertainValue` to `double`, there would be an unnecessary and probably undetected copy. So `auto&&` is correct, although I didn't know that when I made the change.

Isn't C++ fun?

Issue #89, opened 2021-10-07
Test installation and integration into other projects


dabele:
The memilio c++ libraries are intended to be included into other projects using cmake, either by installing memilio first or including it directly as a subdirectory (similar to what memilio does with its dependencies). Both ways should be tested and fixed if necessary.

# Steps for integration with install:
in shell:
```bash
mkdir build && cd build
cmake .. #maybe with options
cmake --build .
cmake --install .
```
then in another project's CMakeList.txt:
```cmake
find_package(memilio REQUIRED) #maybe check specific components?
target_link_libraries(<target> PUBLIC memilio::memilio) #also memilio::secir and other models
```

# Steps for integration with add subdirectory:

in another project's CMakeLists.txt:
```cmake
add_subdirectory('memilio/cpp')
target_link_libraries(<target> PUBLIC memilio::memilio)
```
or 
```cmake
target_link_libraries(<target> PUBLIC memilio)
```

PR #187, opened 2022-01-28
65 use conan package manager

### LINKED ISSUE #65: clang above version 9 cannot build bundled jsoncpp version
Clang versions above 9 cannot build the bundled jsoncpp version due to warnings that are treated as errors. It is (probably) possible to use e.g. clang 12 with your own jsoncpp installation. The clang jobs with version 12 have been temporarily disabled in CI (#64) .

Possible solutions:
- switch to more recent jsoncpp version which probably doesn't have this issue. probably requires code changes and maybe other cmake changes
- find a way to configure the jsoncpp build to ignore warnings
- ?

### LINKED ISSUE #88: CMake deprecation warning in bundled googletest
CMake throws this warning:

```
CMake Deprecation Warning at CMakeLists.txt:4 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


CMake Deprecation Warning at googlemock/CMakeLists.txt:45 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


CMake Deprecation Warning at googletest/CMakeLists.txt:56 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

Upgrading google tests to the newest Verson 1.11.0 will likely fix this error.

### LINKED ISSUE #165: Cmake fails for cmake version 3.10
cmake fails when installing targets defined as subdirectories for cmake version 3.10.

dabele:
## Changes

- Introduce conan package manager, see readme for usage
- removed some cmake options. options like MEMILIO_USE_SYSTEM_BOOST or maybe MEMILIO_USE_CONAN could be reintroduced, but I think simple is better for now. I have experimented with this and have a version of this ready if needed that needs only minor fixes.
- removed obsolete third party things
- new jsoncpp version
- test with clang 12 in CI
- speed up CI by caching dependencies
- moved and fixed cmake install code (#165), installing and using memilio works now, but should be more user friendly, e.g. by making memilio installable with conan (#89 remains open for this)

Closes #65
Closes #165
Closes #88 

### Checks by code author:


dabele:
@reneSchm: To help @mknaranja with the review, please try building our project on this branch. Start with a clean build folder and try to follow the README before asking me or before googling - ask me of course if you get stuck. Also try at least a few of the cmake options we provide. 

Please check if
- everything works, i.e. builds successfully and tests run
- the READMEs are complete and understandable. Let me know if you had any problems, even if you managed to solve them yourself

reneSchm:
- Using a clean install, cmake configures and builds the project without errors, and all tests succed
  - (except for the occasional bad choice of parameters by random ParameterStudies)
- The readme in cpp/ was sufficient to build the project, and could be found via a link from the main readme
  - Additionally, an existing venv was used to install conan

However, by default a local install of boost is preferred over the conan install, without any indication of this behaviour during configure. In my case, this causes the build to fail.
Unloading the PrgEnv with boost resolves the issue. (The preinstalled version of boost fails to declare variables used in the makro BOOST_OUTCOME_TRY, which causes compilation to fail with "undeclared variables".)

Further remarks on conan:
- The first configure step took several minutes, configuring boost for the majority of the time. Without conan, this step is significantly faster (on my machine). Build time appears to be unaffected.
- Conan installed all dependencies in `~/.conan/data/...`. Should we provide an option/default to place these installs inside the project or build directory?

dabele:
Thank you for testing.

> by default a local install of boost is preferred over the conan install, without any indication of this behaviour during configure. In my case, this causes the build to fail. Unloading the PrgEnv with boost resolves the issue. 

This is should not happen. The thirdparty/CMakeLists.txt sets the boost_DIR variable to the build directory, that should be used as a hint by the find_package function. I'll have to investigate that. Which system did you try this on? Do you happen to know if this affected only the headers or the libraries as well? Could you please send me the CMakeCache.txt from the build directory and/or a log of the cmake run? My 
guess would be that cmake finds the intended package and adds the correct include paths, but the boost headers are included in the system include path and are found first by the compiler.

>(The preinstalled version of boost fails to declare variables used in the makro BOOST_OUTCOME_TRY, which causes compilation to fail with "undeclared variables".)

Out of curiosity, even though I think this probably doesn't need to be fixed, which version of boost was this?

>     * The first configure step took several minutes, configuring boost for the majority of the time. Without conan, this step is significantly faster (on my machine). 

This is expected. A new run should be much faster, though, since it doesn't need to rebuild, it reuses the stuff in ~/.conan

> Build time appears to be unaffected.

The build itself should be a bit faster since we don't need to build the dependencies anymore, they are built or downloaded during cmake configure.

>     * Conan installed all dependencies in `~/.conan/data/...`. Should we provide an option/default to place these installs inside the project or build directory?

The default should be the home directory, since that is almost always what you want. You usually don't want to put the data in the build directory, since then you can't use the same dependencies for multiple builds and everything needs to be rebuilt when you clear the build dir, which is necessary sometimes. Conan itself can be configured in multiple different ways to put the data in a different directory. I don't think we should duplicate that, but maybe we should put a note in our README that the directory can be changed, together with a link to the conan documentation?

reneSchm:
> Which system did you try this on?

The TeamServer without any modules or PrgEnv loaded, since my go-to PrgEnv has boost included.

> Do you happen to know if this affected only the headers or the libraries as well?

The headers are affected, since the compiler errors report the paths `/tools/modulesystem/.../<boost header>` when expanding the macros - which I assume is supposed to look something like `~/.conan/.../<boost header>` instead. But compilation stops at 1%, so I do not think any libraries have benn linked yet.

> Out of curiosity, even though I think this probably doesn't need to be fixed, which version of boost was this?

I am using gcc-9.4.0 and boost-1.78.0 (from spack, compiled with gcc-9.4.0).

> The build itself should be a bit faster since we don't need to build the dependencies anymore, they are built or downloaded during cmake configure.

I did not do any timings, so it is likely that I just did not notice any speedup when building. Although, reconfiguring is a lot faster using conan.

> I don't think we should duplicate that, but maybe we should put a note in our README that the directory can be changed, together with a link to the conan documentation?

I think that would by good. If a user were to not notice the install, the dependecies would be left in home/ when removing the project, which may be unwanted.

dabele:
resolved merge conflicts

dabele:
I investigated the issue with boost headers on the teamserver. I was able to reproduce it. It is as I suspected, cmake finds the right package (the one from Conan) but gcc finds the include directory from spack first. 

I don't believe there is an acceptable way to fix this from our side. gcc checks directories in this order:
1. directories specified on the command line with `-I`
2. directories in the CPATH environment variable
3. directories specified on the command line with `-isystem`
4. directories in CPLUS_INCLUDE_PATH env var
5. default paths like /usr/include

spack adds the directories to CPATH. Cmake adds the directories with `-isystem`. This is usually good, as warnings from headers in system include dirs get automatically ignored. Cmake can be told to use `-I` instead of `-isystem`, but of course we would have to deal with warnings from all dependencies. (Also, it feels wrong! :smile:)

The variable that spack adds its directories to can be configured on a per module basis. Maybe this can be fixed on our servers.

dabele:
> > I don't think we should duplicate that, but maybe we should put a note in our README that the directory can be changed, together with a link to the conan documentation?
> 
> I think that would by good. If a user were to not notice the install, the dependecies would be left in home/ when removing the project, which may be unwanted.

Added a paragraph to cpp/thirdparty/README and a short hint in cpp/README

dabele:
> I am using gcc-9.4.0 and boost-1.78.0 (from spack, compiled with gcc-9.4.0).

I found the problem with boost 1.78. BOOST_OUTCOME_TRY(i, expr) does not declare `auto&& i` anymore, it assigns an existing variable. There is a legacy macro OUTCOME21_TRY to restore the old behavior. 

looks like we will have to upgrade to boost 1.78 anyway. visual studio 2022 (e.g. on the windows 2022 github runner) does not support 1.75, see https://github.com/DLR-SC/memilio/runs/5072775783?check_suite_focus=true

xsaschako:
I did a clean install on windows. This always led to an error wheni set up cmake with conan. The error output was about only finding a corrupted boost file. I tried some things which didn't help. Then i tried it on linux through nomachine.

- Cloned memilio and switched on branch 65-reduced-options.
- Created a new virt env through python and upgraded pip to the most recent version.
- Installed conan and cmake as descibed in the readme.
- Ran cmake in the build directory as descibed in the readme. This is were on windows the error occured. It worked fine in linux. (GNU compiler vs Visual studio build tools).
- Build the library with `cmake --build .` Which worked fine.
- `./tests/memilio-test` worked fine.
- `./examples/secir-example` worked fine.

The Readme was very clearly structured and easily to follow. As described above we should discuss the implementation and description on windows.

xsaschako:
A minor nitpick is that the readme code is written for linux. For example `mkdir build && cd ./build`doesn't work on windows.

mknaranja:
I close this for now.

PR #255, opened 2022-04-25
254 batch simulation

### LINKED ISSUE #254: MPI parallelisation of simulations/paper_202011
Enable `cpp/simulations/paper_202011` to be run in parallel using mpi, so that each rank can run its own ParameterStudy.
The results can be collected and evaluated on a single rank, e.g. to calculate percentiles.

Realted to #126

reneSchm:

### Checks by code author:

Closes #254


reneSchm:
Currently, the main feature of running parallel parameter studies using mpi is implemented, but does not yet produce exactly the same output as sequential execution. The outputs of sequential/parallel runs (using the same seed and/or saved graph) only differ in what appears to be rounding errors with magnitude of about 1e-10.

What I examined so far:
- CachedRNG works as expected, given that the cache is large enough. That is, with any given seed the generated numbers are the same as if using the normal RNG sequentially, and at any call to thread_local_rng during computation the same number is returned.
  - I verified the values of the random numbers, and sporadically tested the values of ParameterDistribution::get_sample, which they are usually used for. Both are the same for parallel/sequential.
-  Serialization to json does not cause the "rounding" error, it occurs somewhere before the output. At least the erroneous numbers are serialized correctly.
- The following output is identical between sequential/parallel runs:
  - the saved graphs in Save mode
  - the percentile folders p05 to p95
  - the results of run0 (by default, 0 is not the root rank)
  - all gathered results, i.e. Result.h5, result_runX.h5 and result_runX_sum.h5
    - technically the files differ, since they include the filepath they were written to, but that does not matter
- the number of processes used for parallel runs does not matter (as long as it is more than 1)

Note:
Since #394 caused the Save and Load mode to use different Random numbers, I did not compare their results, only the results of sequential and parallel runs in the same mode. That said, other than the saved graph and having different numerical values, both modes behaved the same in regards to the points above.

mknaranja:
Closed in favor of #581 which built upon this initial implementation.

PR #282, opened 2022-05-11
270 generalize use of analyze result

### LINKED ISSUE #270: Generalize use of analyze_result for different time points and use in ABM
- Make analyze_result usable in ABM.
- Allow for different time points to be interpolated. Right now, interpolation points are fixed as full days.

There is already a way to incorporate multiple time series, but this allows to do this for finer resolution.

DavidKerkmann:
## Changes

Added functionality to interpolate simulation results in form of time series not only at fixed times (days) but at arbitrary time points in between start time and end time of simulation.

### Checks by code author:

closes #270


DavidKerkmann:
We should discuss renaming "interpolate_simulation_result_days" into "interpolate_simulation_result" (which then automatically uses the days as interpolation points).

xsaschako:
> We should discuss renaming "interpolate_simulation_result_days" into "interpolate_simulation_result" (which then automatically uses the days as interpolation points).

In its current implementation i think the names are apropraite. If we have other options,e.g. hours or weeks, this might change.

DavidKerkmann:
I tried to fix the python bindings which failed due to a change in the function names and parameters with the latest commit. When running the install on the py-simulation, it still shows an error with the ```interpolate_ensemble_results``` function, which was not changed in this branch. An install on the current main branch works just fine, so I assume it is related to something we changed, but I cannot figure out what. @dabele  Since I am not familiar with the C++ code regarding the python bindings, could you have a quick look?

dabele:
I can see you mixed up include paths. `#include "models/secir/<xyz>.h"`. The correct path (the one set by the cmake config) would be `"secir/<xyz>.h"`. Try fixing that, I can take a closer look if it doesn't help.

DavidKerkmann:
I reverted all changes regarding the include paths. This had no effects on my build.

cpp/models/secir/analyze_result.h
```diff
@@ -57,7 +70,7 @@ std::vector<InterpolateResultT<T>> interpolate_ensemble_results(const std::vecto
     std::vector<InterpolateResultT<T>> interpolated;
     interpolated.reserve(ensemble_results.size());
     std::transform(ensemble_results.begin(), ensemble_results.end(), std::back_inserter(interpolated), [](auto& run) {
-        return interpolate_simulation_result(run);
+        return interpolate_simulation_result_days(run);
     });
     return interpolated;
 }
```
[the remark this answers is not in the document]
DavidKerkmann: That's right. I didn't consider different arguments when changing this function. I changed the name of the ``ìnterpolate_simulation_result_days```to match with the other one, so they now just differ in arguments. Later, we can for example implement a function that takes a different common time (other than days) and automatically creates the corresponding vector of interpolation points, if needed.

cpp/memilio/math/interpolation.h
```diff
@@ -0,0 +1,45 @@
...
+ * @param[in] x_2 Right node of interpolation.
+ * @param[in] y_1 Value at left node.
+ * @param[in] y_2 Value at right node.
+ * @param[out] unnamed Interpolation result.
+ */
+template<typename X, typename V>
+auto linear_interpolation(const X& x_eval, const X& x_1, const X& x_2, const V& y1, const V& y2)
+{
+    auto weight = (x_eval - x_1) / (x_2 - x_1);
+    return (1-weight)*y1 + weight*y2;
```
dabele, line 40: save one multiplication:
```suggestion
    return y1 + weight * (y2 - y1);
```

cpp/models/secir/analyze_result.cpp
```diff
@@ -18,52 +18,72 @@
...
+    return interpolate_simulation_result(simulation_result, tps, abs_tol);
+}
 
-    auto day = day0;
+TimeSeries<double> interpolate_simulation_result(const TimeSeries<double>& simulation_result, const std::vector<double>& interpolation_times, const double abs_tol)
+{
+    assert(simulation_result.get_num_time_points() > 0 && "TimeSeries must not be empty.");
+    
+    assert((*std::min_element(interpolation_times.begin(), interpolation_times.end()) + abs_tol >= simulation_result.get_time(0) && *std::max_element(interpolation_times.begin(), interpolation_times.end()) - abs_tol <= simulation_result.get_last_time()) && "Interpolation times have lie between simulation times (up to given tolerance).");
+    unused(abs_tol);
```
dabele, line 50: do you really only have the tolerance argument for this assert?

cpp/models/secir/analyze_result.cpp
```diff
@@ -18,52 +18,72 @@
...
+    
+    if ((int)interpolation_times.size() == 0) {
+        std::cout << "Warning: Vector of interpolation times is empty. Returning empty TimeSeries.";
+    }
```
dabele, line 54:
```suggestion
    assert(interpolation_times.size() > 0);
```

cpp/models/secir/analyze_result.cpp
```diff
@@ -18,52 +18,72 @@
...
+    
     TimeSeries<double> interpolated(simulation_result.get_num_elements());
-    interpolated.reserve(day_max - day0 + 1);
-    interpolated.add_time_point(day, simulation_result.get_value(0));
-    day++;
 
-    //interpolate between pair of time points that lie on either side of each integer day
-    for (int i = 0; i < simulation_result.get_num_time_points() - 1;) {
+    int pointer_interp{};
```
dabele, line 58: I have a suspicion that you chose the an intialization that works by accident. But even if it was intentional, please be explicit about zero initialization:
```suggestion
    int pointer_interp = 0;
```

dabele, line 58: please name such a variable `..._index` or `index_...` (or `idx_...`), it's not a pointer.

Also see if you could use a more fitting type to reduce the number of casts, I think `pointer_interp` could be size_t, `pointer_sim` could be Eigen::Index

dabele (changes requested):

cpp/models/secir/analyze_result.h
```diff
@@ -31,22 +31,35 @@ namespace mio
...
+ * @param simulation_result time series to interpolate
+ * @param interpolations_times std::vector of time points at which simulation results are interpolated.
+ * @param abs_tol  absolute tolerance given for doubles t0 and tmax to account for small deviations from whole days. The vector of interpolation times
+ * @return interpolated time series at given interpolation points
+ */
+TimeSeries<double> interpolate_simulation_result(const TimeSeries<double>& simulation_result, const std::vector<double>& interpolation_times, const double abs_tol = 1e-14);
 
 /**
- * helper template, type returned by overload interpolate_simulation_result(T t)
+ * helper template, type returned by overload interpolate_simulation_result_days(T t)
```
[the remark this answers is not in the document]
DavidKerkmann: Yes, I forgot to change this comment.

cpp/models/secir/analyze_result.cpp
DavidKerkmann: Yes, since you pass a vector of the interpolation times, these only have to be checked if the lie between the boundaries of the time series, up to the tolerance. I'm open for suggestions if there is a better solution.

cpp/memilio/math/interpolation.h
DavidKerkmann: Sure, I will change it.

cpp/models/secir/analyze_result.cpp
DavidKerkmann: The idea here was that maybe the function is called with an empty vector (for example in some kind of loop that iterates over specific time points and accumulates, and some call might be with an empty vector). In that case, one should just return an empty TimeSeries and not abort the program. One could also remove this completely and just take a comment in the code. Your thoughts?

cpp/models/secir/analyze_result.cpp
DavidKerkmann: Yes, I will change it.

DavidKerkmann: That's a good point, I changed it.

xsaschako (commented):
Looks good!

cpp/models/secir/analyze_result.cpp
dabele: I assume the algorithm itself is robust to any tolerance (i.e. doesn't fail or crash), or the argument wouldn't be unused. In that case you don't need to check this condition at all.

If you want to check this, it's ok to not use tolerance at all. It's easy for the user to generate values that lie safely within the (closed) interval. There is no need to use tolerances to check values that you haven't generated yourself.

Your check crashes if `interpolation_times` is empty, by the way.

cpp/models/secir/analyze_result.cpp
dabele: If this is an intended use case and the algorithm handles it ok, there doesn't need to be any warning I think. The user would have no way to avoid the warning despite the code being correct. Anyway, if you want to add a warning, use `log_warning` from our logging framework.

DavidKerkmann: Thanks for your insights, I will just remove the warning then and immediately return an empty TimeSeries if the vector is empty.

cpp/models/secir/analyze_result.cpp
DavidKerkmann: It is correct that the user can easily produce values that lie within the interval.
The only problem I have is that the other function that generates interpolation values for the "days" from a given tolerance might allow for values that lie slightly outside of the interval and then calls the function with this vector that does not lie strictly within the interval. Having a check without tolerance would then lead to an assertion error.
On the other hand, having no check would lead to potentially misleading/wrong results when calling the function with a wrong vector by mistake since no error would be generated.

DavidKerkmann: I removed the tolerance in the main analyze_result function and changed the assertions. The function now allows for one time point that is smaller or equal to the first simulation result time and one time point that is larger than the last simulation result time. The values at these time points might not be meaningful and are intended to be used only for small tolerances from the boundaries of the simulation time.

dabele (changes requested):
please also don't forget clang format, line length makes the code hard to read

cpp/models/secir/analyze_result.cpp
DavidKerkmann: Yes, I did. I forgot to use my changes that I had saved somewhere else.

dabele:
> > please also don't forget clang format, line length makes the code hard to read
> 
> @xsaschako Could you maybe format the files? I didn't manage to get clang-format to work on my Mac.

If you can't get it to work in the future, for consistent readability stick to line length, indentation and braces as they are in the rest of the code, the rest isn't that important. that wouldn't be too much work manually and probably can be automated in your editor of choice.

xsaschako:
> > please also don't forget clang format, line length makes the code hard to read
> 
> @xsaschako Could you maybe format the files? I didn't manage to get clang-format to work on my Mac.

dabele (approved):

PR #319, opened 2022-07-13
293 Adjust splitting for default value of out_folder

HenrZu:
The problem here was that my path, in which the memilio folder was located, also contained the word memilio.
Previously, for the default value of `out_folder´, the absolute file path was splitted at the first appearance of 'memilio'.
The change makes it more flexible and allows higher subfolders to be named like "code_memilio" etc.

### Checks by code author:
* [ ] There is at least one issue associated with the pull request.
* [ ] The branch follows the naming conventions as defined in the [git workflow](git-workflow).
* [ ] New code adheres with the [coding guidelines](coding-guidelines)
* [ ] A local test was succesful

### DIFF
--- a/pycode/memilio-epidata/memilio/epidata/defaultDict.py
+++ b/pycode/memilio-epidata/memilio/epidata/defaultDict.py
@@ -33,10 +33,13 @@ import os
 from datetime import date
 
 default_file_path = os.path.dirname(os.path.abspath(__file__))
-dfp_vec = default_file_path.split('memilio')
+# Divide the path at the Memilio subfolder. 
+# This prevents a changed out_folder by previous subfolders containing the name Memilio.
+# Since Windows ('\\') and Linux ('/') use different directory separators, we use os.sep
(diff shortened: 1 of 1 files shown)


dabele:
Tiny improvement and I don't think it makes anything worse, so I'm inclined to approve it. But it doesn't solve the problem.

Scenarios where this still fails or produces unexpected results:
- repository directory not named memilio
- memilio-epidata package installed in python environment that is outside of the repository, e.g. global python environment
- some parent folder of repository also called memilio

All of these are very reasonable, I have personally used all of them at some point. I still think the only proper way to do this is with no default value at all (or at most the current working directory)

HenrZu:
I agree with you that directly removing the default value for the output folder is a better option. Originally I wanted to relate it to another issue and just fix my actual problem.
I have changed the code so that the value is no longer set by default
Thus we should be more independent of the naming in the file path and also of use in installed packages.

xsaschako:
@HenrZu What's needed here?

PR #431, opened 2022-11-07
92 run examples in ci to ensure basic functionality abm

### LINKED ISSUE #92: run examples in CI to ensure basic functionality
C++ examples are built as part of normal CI, but not run. Python examples are not tested at all. 

To ensure at least basic functionality, each example should be run at least once, maybe not in all configurations. I don't think we should check specific behavior (that's what tests are for) but they should at least run without crashing and finish with no error code.

A custom cmake target is probably the most maintainable, at least for c++. then the definiton of examples is close to the test run.

xsaschako:

### Checks by code author:
* [x] The following questions are addressed in the documentation*:  Developers (what did you do?, how can it be maintained?), For users (how to use your work?), For admins (how to install and configure your work?)

* [x] Is the code clean of development artifacts e.g., unnecessary comments, prints, ...
* [x] The ticket goals for each associated issue are reached or problems are clearly addressed (i.e., a new issue was introduced).
* [x] There are appropriate **unit tests** and they pass.
* [x] The git history is clean and linearized for the merge request.
* [x] Coverage report for new code is acceptable. 

closes #92


xsaschako:
Suggested edit:
```diff
diff --git a/.github/actions/linux-run_examples/action.yml b/.github/actions/linux-run_examples/action.yml
index 8fd1e1115..7f6638686 100644
--- a/.github/actions/linux-run_examples/action.yml
+++ b/.github/actions/linux-run_examples/action.yml
@@ -31,8 +31,7 @@ runs:
         sudo chmod -R a+x ./
         numberoffiles=$(ls -dq *_example | wc -l)
         echo $numberoffiles
-        if [$numberoffiles -gt 0 ]
-        then 
+        if [$numberoffiles > 0 ] then;
           echo "Running Linux examples."
           find ./ -name "*_example" -type f -exec {} \;
         else
diff --git a/.github/actions/windows-run_examples/action.yml b/.github/actions/windows-run_examples/action.yml
index 44b29cf8c..fbc0b74a1 100644
--- a/.github/actions/windows-run_examples/action.yml
+++ b/.github/actions/windows-run_examples/action.yml
@@ -18,7 +18,7 @@ runs:
         cd cpp/build-win/bin
         dir
         $errchk = (Get-ChildItem -Path ./| Where-Object { $_.Name -like "*_example.exe"} | Measure-Object).Count
-        if( $errchk -gt 0){
+        if($errchk -gt 0){
           Write-Output "Running windows Examples."
           Get-ChildItem -Path ./| Where-Object { $_.Name -like "*_example.exe"}
           Get-ChildItem -Path ./| Where-Object { $_.Name -like "*_example.exe"} | Foreach {Start-Process $_.FullName}

```

xsaschako:
@mknaranja 
This runs all example files (files with names _example in example folder) and returns a fail if something fails to execute. 
One issue: These examples must at least return error code 0 when called just by name.
This is the case for everything except the example associated with ode_secir_read_graph.cpp. There you have to call it with a file. This is a little cumbersome to implement in CI, so for now, I have this suboptimal fix to return 0 instead of -1. One could argue that an example is runnable with just the name for demonstration purpose, so this return code could make sense. Otherwise, there are a few options: 
- don't run it in these example runs.
- run it with an argument (cumbersome to implement).
- rename it, so it hasn't the “_example” in the name.

Also FYI, there were some other implementation path's I could've taken, which could be better/worse:

As Daniel suggested, we could have called these examples at build time in CMake. This has two drawbacks:
1. The fails would have been in the Build actions and thus not at a first glance recognizable as runtime fails, which happen in the examples.
2. There may also have been a nice way to call all example executables, but as far as I know, you would've to add the names of each executable, so one would have to add new examples manually. I think this would've been easily forgotten. With this approach, this isn't the case as long as the name has the _example ending, which it should have anyway for consistency.

DavidKerkmann (commented):
Thanks for adding this and for the nice discussion today!
I have two small issues to think about:
- For the use of examples with arguments, as in `ode_secir_read_graph.cpp`, there could be a handling for providing no arguments. In this case, the example is simply terminated right now. Instead, one could provide a default path to a default graph to read in. The folder  saved in `MEMILIO_DATA_DIR` doesn't exist in the current repository. @mknaranja Is there an option to add some default data to read in to make this test execution meaningful?
- For error finding, it would be good to add individual outputs to the runs of the examples,  for example: `Running abm_example`, `Running ode_seir_example`, etc. Would that be possible?

xsaschako:
Internally, it doesn't run parallel, it just doesn't wait for the program to be finished and just launches the next.
In Microsoft, it does. That isn't perfect, but I don't see an elegant solution right now in bash.

xsaschako:
> Thank you for adding the items. This looks very good! I have one question. The output might look like this:
> 
> `22.10000000000000 8726.27578087024995 191.09707482138006 113.16338926385502 249.95754471252141 58.36880681769578 8.43750112935309 646.10349614940537 6.59640623553666 (https://github.com/DLR-SC/memilio/actions/runs/4307512029/jobs/7513135808#step:3:248) bash -c ./ode_secir_read_graph_example (https://github.com/DLR-SC/memilio/actions/runs/4307512029/jobs/7513135808#step:3:249) 23.10000000000000 8661.01365844235443 195.88933741032318 116.27202175741527 258.33730584078597 60.85775248903810 8.79297944359873 691.87656846422817 6.96037615225292`
> 
> It looks like the examples are run in parallel so that the output is shuffled in an undefined behaviour and makes it hard to see the output for just one example run. I don't know if we should change this or not, up for discussion. Otherwise, this is ready to be merged.

I don't know what causes this, it is just that one time as far as I can see. The others look fine and therefore I think this is okay even though not 100% perfect.

DavidKerkmann (approved):

mknaranja:
@xsaschako Could you resolve conflicts and @DavidKerkmann could you merge then? I'm okay with it if you approved it.
Thank you both.

DavidKerkmann (approved):

PR #536, opened 2023-01-19
467 use uncertain values for abm parameters

### LINKED ISSUE #467: Use uncertain values for ABM parameters
In order to allow parameter studies, ABM parameters should not be double values but uncertain. Here, we can use the UncertainValue object as introduced for the ODE model and a uniform distribution in first place.

khoanguyen-dev:

### Checks by code author:
Always to be checked:
* [x] Make sure that the [pre-commit linting/style checks pass](https://github.com/DLR-SC/memilio/wiki).

If functions were changed or functionality was added:

If new functionality was added:

If new third party software is used:
* [ ] Did you pay attention to its license? Please remember to add it to the wiki after successful merging.

If new mathematical methods or epidemiological terms are used:
* [ ] Are new methods referenced? Did you provide further documentation? Has the glossary been updated? 

The following questions are addressed in the documentation if need be: 
* [ ] Developers (what did you do?, how can it be maintained?)
* [ ] For users (how to use your work?)
* [ ] For admins (how to install and configure your work?)

* [x] Is the code clean of development artifacts e.g., unnecessary comments, prints, ...
* [x] The ticket goals for each associated issue are reached or problems are clearly addressed (i.e., a new issue was introduced).
* [x] There are appropriate **unit tests** and they pass.
* [x] The git history is clean and linearized for the merge request. All reviewers should squash commits and write a simple and meaningful commit message.
* [x] Coverage report for new code is acceptable. 

Closes #467


cpp/simulations/abm.cpp
```diff
@@ -22,20 +22,35 @@
...
-mio::abm::InfectionState determine_infection_state(double exposed, double infected, double carrier, double recovered)
+mio::abm::InfectionState determine_infection_state(mio::UncertainValue exposed, mio::UncertainValue infected,
+                                                   mio::UncertainValue carrier, mio::UncertainValue recovered)
 {
-    double susceptible          = 1 - exposed - infected - carrier - recovered;
-    std::vector<double> weights = {susceptible,  exposed,      carrier,       infected / 3,
-                                   infected / 3, infected / 3, recovered / 2, recovered / 2};
+    mio::UncertainValue susceptible = 1 - exposed - infected - carrier - recovered;
+    std::vector<ScalarType> weights = {susceptible,  exposed,      carrier,       infected / 3,
+                                       infected / 3, infected / 3, recovered / 2, recovered / 2};
```
DavidKerkmann, line 53: I don't see the point in changing it to UncertainValues here as the function only performs simple arithmetic on some values.

cpp/simulations/abm.cpp
```diff
@@ -301,8 +316,10 @@ void create_assign_locations(mio::abm::World& world)
     auto testing_min_time = mio::abm::days(2);
     auto start_date       = mio::abm::TimePoint(0);
     auto end_date         = mio::abm::TimePoint(0) + mio::abm::days(60);
-    auto probability      = 1.0;
-    auto test_type        = mio::abm::AntigenTest();
+
+    auto probability = mio::UncertainValue();
+    assign_uniform_distribution(probability, 1.0, 1.0);
+    auto test_type = mio::abm::AntigenTest();
```
DavidKerkmann, line 322: This doesn't change anything in the functionality.
If you want to use an uncertainty in the probability, then change the distribution to a non-constant distribution and draw from it when creating the testing scheme. This will then result in an uncertain initial parameter for this simulation, but will be constant at runtime.

cpp/simulations/abm.cpp
```diff
@@ -401,27 +418,25 @@ void create_assign_locations(mio::abm::World& world)
         std::vector<mio::abm::TestingCriteria>{mio::abm::TestingCriteria({}, test_at_school, {})};
 
     testing_min_time           = mio::abm::days(7);
-    probability                = 1.0;
     auto testing_scheme_school = mio::abm::TestingScheme(testing_criteria_school, testing_min_time, start_date,
-                                                         end_date, test_type, probability);
+                                                         end_date, test_type, mio::UncertainValue(1.0));
```
DavidKerkmann, line 422: Again, this doesn't do anything.

cpp/simulations/abm.cpp
```diff
@@ -401,27 +418,25 @@ void create_assign_locations(mio::abm::World& world)
...
     auto testing_criteria_work =
         std::vector<mio::abm::TestingCriteria>{mio::abm::TestingCriteria({}, test_at_work, {})};
 
-    testing_min_time = mio::abm::days(1);
-    probability      = 0.5;
-    auto testing_scheme_work =
-        mio::abm::TestingScheme(testing_criteria_work, testing_min_time, start_date, end_date, test_type, probability);
+    testing_min_time         = mio::abm::days(1);
+    auto testing_scheme_work = mio::abm::TestingScheme(testing_criteria_work, testing_min_time, start_date, end_date,
+                                                       test_type, mio::UncertainValue(0.5));
```
DavidKerkmann, line 431: See above.

cpp/simulations/abm.cpp
```diff
@@ -401,27 +418,25 @@ void create_assign_locations(mio::abm::World& world)
...
     world.get_testing_strategy().add_testing_scheme(testing_scheme_work);
 }
 
 /**
  * Assign an infection state to each person.
  */
-void assign_infection_state(mio::abm::World& world, double exposed_pct, double infected_pct, double carrier_pct,
-                            double recovered_pct)
+void assign_infection_state(mio::abm::World& world, mio::UncertainValue exposed_pct, mio::UncertainValue infected_pct,
+                            mio::UncertainValue carrier_pct, mio::UncertainValue recovered_pct)
```
DavidKerkmann, line 439: Same as above.

DavidKerkmann (changes requested):
Thanks for changing the parameters to UncertainValues.
However, none of the functionality of UncertainValues is actually used right now.
There are two options to use UncertainValues:
1. One can draw from the given distribution when setting up the problem. For example in 
`infection_params.get<mio::abm::SusceptibleToExposedByCarrier>()[{mio::abm::AgeGroup::Age0to4,
                                                                     mio::abm::VaccinationState::Unvaccinated}]  = 0.05;`
you could add a distribution and then draw the (fixed) value that will be used in the simulation from the distribution. Of course, this doesn't need to happen in this particular simulation, so one can just leave it as it is here.
2. The values of the distribution could be drawn each time (or at some particular times) a value is needed. For example, the test parameters in `if (random < params.sensitivity) {`and so on in person.cpp could be drawn from a distribution each time. This option should be implemented strictly for those parameters where this makes sense. If the option is not used, a fallback to a fixed value is already implemented (see uncertain_value.cpp).

We should discuss in the group which parameters should have the option to be drawn multiple times, and not just in the initialisation of the simulation, such that the functionality can be changed. To take the probability of testing in a TestingScheme as an example: If we want to be able to have TestingSchemes where the probability is not fixed, but is drawn from a distribution, then the double in the header has to be changed to an UncertainValue und the evaluation to a draw_sample.

Also,

`m_random_workgroup        = UniformDistribution<double>::get_instance()();`

and related functions should be changed to draw from a new parameter that states a distribution of possible times, I would suggest.

khoanguyen-dev:
The list of input parameters for a simulation can be sorted as followed:

1/ To be drawn from distribution once during the simulation initialisation (or set as a constant):

- Parameters::Global Infection params <IncubationPeriod, SusceptibleToExposedByCarrier, SusceptibleToExposedByInfected, CarrierToInfected, CarrierToRecovered, InfectedToRecovered, InfectedToSevere, SevereToCritical, SevereToRecovered, CriticalToDead, CriticalToRecovered, RecoveredToSusceptible, DetectInfection, MaskProtection> for each age group and vaccination state
- Parameters::MigrationParameters <LockdownDate, SocialEventRate, BasicShoppingRate, WorkRatio, SchoolRatio, GotoWorkTimeMinimum,GotoWorkTimeMaximum, GotoSchoolTimeMinimum, GotoSchoolTimeMaximum>
- Test_strategy::m_probability
- Person::m_random_workgroup, m_random_schoolgroup, m_random_goto_work_hour, m_random_goto_school_hour
- Person::m_mask_compliance
- Household::pick_age_group_from_age_distribution

2/ To be drawn (multiple times) from distribution while running:

- Person::wear_mask
- Person::get_tested
- Random_event::random_transition
- Test_strategy::run_scheme

cpp/models/abm/parameters.h
```diff
@@ -230,13 +231,13 @@ struct MaximumContacts {
...
-    double sensitivity;
-    double specificity;
+    UncertainValue sensitivity;
+    UncertainValue specificity;
 };
 
 struct GenericTest {
     using Type = TestParameters;
-    static constexpr Type get_default()
+    static Type get_default()
```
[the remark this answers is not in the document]
khoanguyen-dev: It was because constexpr cannot accept non-literal type. I switch them back since I think both sensitivity and specificity of the test can be specified at the beginning of the simulation and they don't necessary have to be UncertainValue.

cpp/simulations/abm.cpp
khoanguyen-dev: It is fixed!

cpp/simulations/abm.cpp
khoanguyen-dev: I change it to have a probability between 0.5-1 and can be drawn at the beginning of each simulation.

cpp/simulations/abm.cpp
khoanguyen-dev: Same as above.

cpp/simulations/abm.cpp
khoanguyen-dev: It is fixed!

cpp/simulations/abm.cpp
khoanguyen-dev: I change it to have a probability between 0.1-0.5 and can be drawn at the beginning of each simulation.

cpp/models/abm/parameters.h
```diff
@@ -285,10 +286,10 @@ struct LockdownDate {
     }
 };
 struct BasicShoppingRate {
-    using Type = CustomIndexArray<double, AgeGroup>;
+    using Type = CustomIndexArray<UncertainValue, AgeGroup>;
     static auto get_default()
     {
-        return CustomIndexArray<double, AgeGroup>(AgeGroup::Count, 1.0);
+        return CustomIndexArray<UncertainValue, AgeGroup>(AgeGroup::Count, 1.0);
```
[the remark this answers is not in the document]
khoanguyen-dev: I still don't get why it would create an error. Could you give more detail?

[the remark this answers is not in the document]
DavidKerkmann: Sorry, I misread something here. Don't think it would create an error. I would suggest to use Type though: `Type({AgeGroup::Count}, 1.0);`

cpp/models/abm/parameters.h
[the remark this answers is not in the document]
DavidKerkmann: I see, yes. However, I think there might be some uncertainty in these values, so I would leave them as UncertainValues and remove the constexpr as you did before. Sorry for the back and forth.

DavidKerkmann (approved):
Looks good to me, only small things.

DavidKerkmann:
> The list of input parameters for a simulation can be sorted as followed:
> 
> 1/ To be drawn from distribution once during the simulation initialisation (or set as a constant):
> 
> * Parameters::Global Infection params <IncubationPeriod, SusceptibleToExposedByCarrier, SusceptibleToExposedByInfected, CarrierToInfected, CarrierToRecovered, InfectedToRecovered, InfectedToSevere, SevereToCritical, SevereToRecovered, CriticalToDead, CriticalToRecovered, RecoveredToSusceptible, DetectInfection, MaskProtection> for each age group and vaccination state

Many of these parameters should be distributions to draw from during the simulation. The parameters of these distributions might have some uncertainty attached to them, too. As discussed in the group, this results in two levels of uncertainty. One could define the parameters for the distributions as level 1 uncertainty and put the corresponding distributions to level 2.

> * Parameters::MigrationParameters <LockdownDate, SocialEventRate, BasicShoppingRate, WorkRatio, SchoolRatio, GotoWorkTimeMinimum,GotoWorkTimeMaximum, GotoSchoolTimeMinimum, GotoSchoolTimeMaximum>

I agree with these. LockdownData will probably be reworked in the future. Here you can see the parameters as level 1 uncertainty, while distributions for the work/school time are implemented explicitly right now.

> * Test_strategy::m_probability

Depends on the test, but I wouldn't consider this to be in the list of "global" parameters.

> * Person::m_random_workgroup, m_random_schoolgroup, m_random_goto_work_hour, m_random_goto_school_hour

These is the level 2 uncertainty to the parameters above. Implemented explicitly right now. We can consider changing these to be UncertainValues/Parameters, too.

> * Person::m_mask_compliance

Agreed, but there should be a parameter added. One could in theory also discuss about this being a level 2 parameter if wearing a mask depends on dynamic items.

> * Household::pick_age_group_from_age_distribution

This only implements a discrete distribution and doesn't belong to the set of parameters.

I would add test parameters here.
 
> 
> 2/ To be drawn (multiple times) from distribution while running:
> 
> * Person::wear_mask

This is a helper bool. Mask compliance is the relevant parameter.

> * Person::get_tested

This is just a function to test a person. The test parameters are handled in level 1.

> * Random_event::random_transition

Same, just a function.

> * Test_strategy::run_scheme

Same, just a function.

@xsaschako @mknaranja Please add your comments when you find the time.

cpp/models/abm/parameters.h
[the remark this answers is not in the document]
khoanguyen-dev: No problem. It's fixed!

cpp/simulations/abm.cpp
khoanguyen-dev: It's done!

cpp/models/abm/parameters.h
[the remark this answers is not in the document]
khoanguyen-dev: It's done!

mknaranja (approved):

Issue #635, opened 2023-05-11
Implement Frontend example for abm output object


jubicker:
### Motivation / Current Behaviour

To test the abm output object we need to implement an example output.

### Enhancement description

The output should be a txt file and have one line per location. It should contain the following columns:

- location Id
- number of output timesteps
- timestep 1 (t1)
- number of agents in location in t1
- agent id 1 (a1)
- time since transmission for a1
- agent id 2 (a2)
- time since transmission for a2
- ...
- timestep 2 (t2)
- number of agents in location in t2
- agent id 1
- time since transmission for a1
- agent id 2
- time since transmission for a2
- ...
- timestep 3 (t3)
- ...

As we need the time since transmission for every agent at every timestep, this has to be added as an attribute to the person class. For agents that are not infected this should be set to -1 in the txt file.

### Checklist

- [X] Attached labels, especially loc:: or model:: labels.
- [X] Linked to project

xsaschako:
This issue will also be handling the output object for frontend ESID

xsaschako:
This issue will just handle the example for the frontend. 
The example for the Demonstrator will be done in #604

xsaschako:
As per Discussion with Moritz:

# ABM Object 23-07-12

#Revisit  
Participants: [[Sascha Korf (DLR-SC-HPC)]]
Tags: [[PANDEMOS]]

---
- no progress due to vacations
  - will be worked on this week
- probably hdf5 files (but open in theory)

Data contained in Output (for single run):
- Locations Lookup
  | Field      | Description                             |
  |------------|-----------------------------------------|
  | LocationID | ID used in the movement vectors         |
  | Latitude   | Latitude of the location's coordinates  |
  | Longitude  | Longitude of the location's coordinates |

- Agent Lookup(?)
  | Field    | Description                                                                               |
  |----------|-------------------------------------------------------------------------------------------|
  | AgentID  | ID of the Agent                                                                           |
  | (HomeID) | Location ID of the Agents home/household                                                  |
  | Groups   | List of stratification groups (socio-economic) the Agent belongs to (i.e. age16-30, male) | 

- Movement Data
  | Field          | Description                                                                           |
  |----------------|---------------------------------------------------------------------------------------|
  | TripID         | ID of the trip/movement vector                                                        |
  | AgentID        | ID of the Agent doing the trip                                                        |
  | (HomeID)       | ID of the Household (or Location ID of the agents home?)                              |
  |                |                                                                                       |
[...]

PR #690, opened 2023-07-11
635 implement example for abm output object and time since transmission variable

xsaschako:
## Changes

Please briefly list the changes made:

- Add Output object for frontend

### Checks by code author

- [ ] New code adheres to [coding guidelines](https://github.com/DLR-SC/memilio/wiki/Coding-guidelines)

- [ ] Appropriate **unit tests** have been added, CI passes and code coverage is acceptable (did not decrease)

### Additional context


cpp/models/abm/world.cpp
```diff
@@ -85,49 +87,70 @@ void World::migration(TimePoint t, TimeSpan dt)
...
+                auto& target_location = find_location(target_type, *person);
+                auto current_location = person->get_location();
+                if (m_testing_strategy.run_strategy(*person, target_location, t)) {
+                    if (target_location != current_location &&
+                        target_location.get_number_persons() < target_location.get_capacity().persons) {
+                        bool wears_mask = person->apply_mask_intervention(target_location);
+                        if (wears_mask) {
+                            m_movement_data.push_back({person->get_person_id(), current_location.get_index(),
+                                                       target_location.get_index(), t, t + dt,
+                                                       mio::abm::TransportMode::Unknown, mio::abm::UnknownActivity,
```
DavidKerkmann, line 105: TransportMode and Activity will be set later I guess once we have implemented a traffic model, but you could set Activity here (or better in the loggers) based on the type of location they go to.

cpp/models/abm/world.cpp
```diff
@@ -85,49 +87,70 @@ void World::migration(TimePoint t, TimeSpan dt)
...
+                                               target_location.get_index(), t, t + dt, trip.trip_mode,
+                                               trip.activity_type, person->get_infection_state(t)});
                     person->migrate_to(target_location);
+                    migrations++;
                 }
             }
             m_trip_list.increase_index();
         }
     }
+    printf("%d migrations\n", migrations);
```
DavidKerkmann, line 145: Please remove this and the counter in the final version.

cpp/models/abm/common_abm_loggers.h
```diff
@@ -0,0 +1,63 @@
...
+        }
+        return person_information;
+    }
+};
+
+struct LogMovementData : mio::LogAlways {
+    using Type = std::vector<mio::abm::movement_data>;
+    static Type log(const mio::abm::Simulation& sim)
+    {
+        return sim.get_world().get_movement_data();
```
DavidKerkmann, line 59: Should be changed, see comment for `m_movement_data`.

DavidKerkmann (changes requested):
Thanks for the implementation. Since you merged with the Braunschweig data set, I didn't go through all the changes from that. I mainly have one concern regarding the determination of the movements in the world rather than the loggers. It might be a good idea to have a smaller test setup, independent of the Braunschweig data set, to check for correct structure of the output. Also, there should be some tests for the loggers, which can be in addition to the smaller test setup.

cpp/models/abm/common_abm_loggers.h
```diff
@@ -0,0 +1,182 @@
...
+
+        Eigen::VectorXd sum = Eigen::VectorXd::Zero(Eigen::Index(mio::abm::InfectionState::Count));
+        PRAGMA_OMP(for)
+        for (auto i = size_t(0); i < sim.get_world().get_locations().size(); ++i) {
+            auto&& location = sim.get_world().get_locations()[i];
+            sum += location.get_subpopulations().get_last_value().cast<ScalarType>();
+        }
+        return std::make_pair(sim.get_time(), sum);
+    }
+};
```
DavidKerkmann, line 146: Since we have this now, we should delete the `TimeSeries m_subpopulations` in the locations and all related functions, too, and directly implement here what is in `Location::store_subpopulations` (or do this in a function `Location::get_current_subpopulations`, which would probably be better). This entry is very large and was only there to save the results for the final entry in the simulation subpopulation TimeSeries, which you now correctly deleted.

DavidKerkmann (changes requested):
Thank you for putting this together, this significantly improves the code and the whole IO process! I have only idea for improvement and one request for change which is not a big task.

Also, I did not see the example for the frontend for #635. I can see that there is a lot included in the simulations, but as far as I understand #635, we wanted to have a setup that just produces a very small data set that we can forward to the frontend to help with development.

cpp/memilio/io/history.h
```diff
@@ -43,21 +43,31 @@ constexpr size_t index_templ_pack()
...
+* LogIf is only logging to Writer if the should_log is true. This needs to be implemented, e.g. for logging only the last step.
+*/
+struct LogOnceStart {
 };
 
 struct LogAlways {
 };
 
+template <typename Cond>
+struct LogIf {
```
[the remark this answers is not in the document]
reneSchm: Thats a good point. The reason we did not remove LogOnce is that it needs a cached value, i.e. a bool indicating whether 'once' already happened. This is stored in the History object, since we wanted to have static loggers, so that we only need to initialize the data, and not worry about managing logger objects.

I think we can probably drop the static requirement, and instead only require that a class logger has a log and should_log function, and is default constructible/destructible. Then the history can keep an additional tuple of loggers, and just call the member functions (i.e. `logger_instance.log()` instead of `Logger::log()` ).
This also allows the loggers to allocate (static) variables, and removes the requirement to inherit from one of the Log* structs (but we can still provide them as examples/defaults like you mentioned).

We could alternatively implement a caching mechanism, but I think non-static loggers is a more future-proof approach.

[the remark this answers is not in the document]
reneSchm: We can also add a function `History::log<Logger>(const T& t)` to allow manually using a single logger. It is trivial to implement, but do we actually need or want it?

cpp/models/abm/person.h
```diff
@@ -155,10 +156,22 @@ class Person
 
     /** 
      * @brief Migrate to a different Location.
-     * @param[in, out] loc_new The new Location of the Person.
+     * @param[in] loc_new The new Location of the Person.
```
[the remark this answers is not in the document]
xsaschako: True, changed it back

cpp/models/abm/common_abm_loggers.h
xsaschako: FYI there exists a function size_t Location::get_subpopulation(TimePoint t, InfectionState state) const which probably does what you mean. I kept that.

cpp/tests/test_history.cpp
```diff
@@ -24,21 +25,25 @@ struct LogAOnce : mio::LogOnce {
     }
 };
 
-template <class... Loggers>
-struct Dummy {
-    template <class T>
-    size_t type_index()
+struct LogStepIf {
+    using Type = int;
+    static bool log(const example& ex)
```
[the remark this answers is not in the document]
xsaschako: good catch, it be undone

cpp/memilio/io/history.h
```diff
@@ -120,50 +102,59 @@ class History
...
+     * @param[in] logger A Logger instance.
+     * @tparam Logger A logger from the list Loggers.
+     */
+    template <class T, class Logger>
+    void log_impl(const T& t, Logger& logger)
     {
-        if (!logged) {
-            WriteWrapper::template log_this<logger>(logger::log(t), m_data);
+        if (logger.should_log(t)) {
+            WriteWrapper::template add_record<Logger>(logger.log(t), m_data);
```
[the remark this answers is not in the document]
reneSchm: We added a short comment to the Writer. The main documentation for these requirements is in the History class and io/Readme.

> Also, can we somehow extract the std::get<mio::index_of_type_v<Logger, Loggers...>>(data) from the writer

Yes, technically, but I don't think we should. The problem is, that the Data tuple only stores the types Logger::Type, which are in general not unique. So we cannot use `std::get<Logger::Type>` or anything similar, we have to look up the `Logger`s index in `Loggers...`. We can change around how we do it, but we need either those two templates, or (somehow) the index of Logger, so that we can call `std::get<size_t>`. We can for example store `Loggers...` in a TypeList, or add a member function to the Writer (or History, which would have its own Problems) to replace the get call, but that would only move what we already do around a bit, which makes the get<index<...>> stuff even more obscure.

If we require that all Data types have to be tuples of length sizeof...(Loggers), then we can add History::get_log<Logger> back (or something similar for the writer to use). But that is a big restriction, it would bar us from e.g. using .

cpp/models/abm/common_abm_loggers.h
```diff
@@ -0,0 +1,238 @@
...
+    case mio::abm::LocationType::ICU:
+        return mio::abm::ActivityType::OtherActivity;
+    case mio::abm::LocationType::Hospital:
+        return mio::abm::ActivityType::OtherActivity;
+    case mio::abm::LocationType::Cemetery:
+        return mio::abm::ActivityType::OtherActivity;
+    default:
+        return mio::abm::ActivityType::UnknownActivity;
+    }
+}
```
reneSchm, line 67: Please add documentation for this struct and function.

cpp/simulations/abm.cpp
```diff
@@ -695,10 +696,13 @@ mio::IOResult<void> run(const fs::path& result_dir, size_t num_runs, bool save_s
     while (run_idx <= num_runs) {
         // Make a simulation using a copy from the original world
         auto sim = mio::abm::Simulation(t0, mio::abm::World(world));
+        // Add a time series writer to the simulation
+        mio::History<mio::abm::TimeSeriesWriter, mio::abm::LogInfectionState> historyTimeSeries{
+            Eigen::Index(mio::abm::InfectionState::Count)};
         // Advance the world to tmax
         sim.advance(tmax);
```
reneSchm, line 703: Should this history be passed to advance?

cpp/tests/test_abm_simulation.cpp
```diff
@@ -103,16 +63,75 @@ TEST(TestSimulation, getWorldAndTimeConst)
...
+    trip_list.add_trip(trip7);
+
+    mio::History<mio::DataWriterToMemory, mio::abm::LogLocationInformation, mio::abm::LogPersonInformation,
+                 mio::abm::LogDataForMovement>
+        historyPersonInf;
+    mio::History<mio::abm::TimeSeriesWriter, mio::abm::LogInfectionState> historyTimeSeries{
+        Eigen::Index(mio::abm::InfectionState::Count)};
+    mio::History<mio::abm::DataWriterToMemoryDelta, mio::abm::LogDataForMovement> historyPersonInfDelta;
+    auto sim = mio::abm::Simulation(mio::abm::TimePoint(0), std::move(world));
+    sim.advance(mio::abm::TimePoint(0) + mio::abm::hours(24), historyPersonInf, historyTimeSeries,
```
reneSchm, line 121: Does this test depend on RNG? If so, please add mocks for all used distributions.
If not, could you add some comments what this test tries to do?

cpp/memilio/io/history.h
```diff
@@ -17,75 +17,79 @@
...
+    {
+    }
+
+    History(typename Loggers::Type... args)
+        : m_data(std::tie(args...))
+    {
+    }
+
     template <class T>
     void log(const T& t)
```
reneSchm, line 90: Please add documentation

xsaschako: Did that

cpp/simulations/abm.cpp
xsaschako: That is correct.

pycode/memilio-simulation/memilio/simulation/abm.cpp
```diff
@@ -218,7 +218,6 @@ PYBIND11_MODULE(_simulation_abm, m)
         .def("advance",
              static_cast<void (mio::abm::Simulation::*)(mio::abm::TimePoint)>(&mio::abm::Simulation::advance),
              py::arg("tmax"))
-        .def_property_readonly("result", &mio::abm::Simulation::get_result)
```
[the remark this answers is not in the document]
xsaschako: For now, we lose that, as far as I can tell. We should issue that (which i have in #861), as it should be quite a change i don't see in this issue.

cpp/tests/test_abm_simulation.cpp
xsaschako: No, all things tested are deterministic

PR #701, opened 2023-07-14
335 force shared libs for python bindings to fix set_log_level

### LINKED ISSUE #335: Set LogLevel not working in Python bindings
`mio.set_log_level(mio.LogLevel.Off)´ is not working.
When using, there are still some info logs showing up.

I'm using the `secir_simply.py´ example.

dabele:
## Changes

The problem to solve: If there are multiple shared objects/DLLs (either multiple python modules or one python module and the memilio libraries themselves) and spdlog is linked statically, there are multiple instances of the log level in each shared object. The set_log_level function in the memilio.simulation python bindings sets the log level in memilio.simulation shared object, but not in the memilio.simulation.secir shared object. So set_log_level doesn't work for secir simulations (or abm simulations or any other models that have their own shared object).

The quickest solution implemented here:
Set MEMILIO_BUILD_SHARED_LIBS=ON in setup.py to build memilio and dependencies as shared libraries when building python bindings so that there is exactly one stored instance of the log level. 
The log level is then stored exactly once in spdlog.so/dll and all logging references that.

Discussions of the drawbacks and alternative solutions and non-solutions below.

### Checks by code author

### Additional context

Closes #335

### DIFF
--- a/pycode/memilio-simulation/setup.py
+++ b/pycode/memilio-simulation/setup.py
@@ -23,6 +23,7 @@ setup(
     packages=find_packages(
         where=os.path.dirname(os.path.abspath(__file__))),
     setup_requires=['cmake'],
+    cmake_args=['-DMEMILIO_BUILD_SHARED_LIBS:BOOL=ON'], #need shared libs so there is one shared log level
     install_requires=[],
     extras_require={'dev': ['numpy >= 1.22'], },
     long_description='', test_suite='memilio.simulation_test',)
--- a/pycode/memilio-simulation/setup.py
+++ b/pycode/memilio-simulation/setup.py
@@ -23,7 +23,8 @@ setup(
     packages=find_packages(
         where=os.path.dirname(os.path.abspath(__file__))),
     setup_requires=['cmake'],
-    cmake_args=['-DMEMILIO_BUILD_SHARED_LIBS:BOOL=ON'], #need shared libs so there is one shared log level
+    # need shared libs so there is one shared log level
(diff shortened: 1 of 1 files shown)


dabele:
This is a tricky problem. I have not found a very nice solution yet.

Possible solutions and their problems:
1. Solution implemented here, set MEMILIO_BUILD_SHARED_LIBS in setup.py, see above. Quickest solution, but problems: 
    - cmake flag could in theory be overwritten by users
    - forces shared objects, which may not be desired?
    - breaks if the build process changes (e.g. package manager) or user supplies his own static spdlog, basically anytime when the spdlog build is not controlled by us.
2. Putting all python bindings in 1 module as it was before. 
    - a bit more work than the other solution, but manageable
    - but lose the benefits of multiple python binding modules (compile times/parallelization, organization)
    - only works if memilio and spdlog are either both linked statically or both linked dynamically, otherwise could again be multiple set_log_level instances. So also breaks when the build process changes, only slight improvement over the first solution
3. Memilio has its own log level
    - still need to force a shared library build, but a bit more reliable than the first solution, since we don't rely on spdlog being a shared object, it's enough that memilio is a shared object
    - confusing duplication of code from spdlog, i.e., our `log("fmt", ...)` function checks our own log level and ignores the spdlog log level
4. ???

A combination of 2 and 3 would fix the problem reliably as far as I can tell. memilio stores the log_level. If memilio is a shared library, it contains the only log level. If memilio is static, it gets linked into exactly one python binding module, so there is again only one log level.

PR #702, opened 2023-07-17
675 alternative computation of compartments in IDE model

### LINKED ISSUE #675: Implement alternative numerical scheme for compartments of IDE model

annawendler:
## Changes
- In the simulation, we update the compartment sizes now using the flows calculated for the current time step.
-  In the initialization, we compute the compartment sizes using the sum formula from the flows as before.
- The order of the member functions is changed to clarify which functions are used to compute the initial size of the compartments and which functions are used for later time steps.

### Checks by code author

### Additional context

Closes #675


lenaploetzke:
The pipeline currently failed only because of the lower patch coverage due to added lines. When PR #923 is merged, the code coverage will be at 100% also with the changes of this PR.

cpp/models/ide_secir/model.cpp
```diff
@@ -195,49 +250,93 @@ void Model::compute_flow(int idx_InfectionTransitions, Eigen::Index idx_Incoming
...
-void Model::compute_deaths()
+void Model::update_compartments_current_timestep()
 {
     Eigen::Index num_time_points = m_populations.get_num_time_points();
 
+    // Exposed
+    m_populations.get_last_value()[Eigen::Index(InfectionState::Exposed)] =
+        m_populations[num_time_points - 2][Eigen::Index(InfectionState::Exposed)] +
+        m_transitions.get_last_value()[Eigen::Index(InfectionTransition::SusceptibleToExposed)] -
+        m_transitions.get_last_value()[Eigen::Index(InfectionTransition::ExposedToInfectedNoSymptoms)];
```
mknaranja, line 295: Quick thought: Wouldn't it be more concise to have a function `update_compartment_from_flow(Eigen::Index idx_InfectionState, std::vector<Eigen::Index> idx_IncomingFlows, std::vector<Eigen::Index> idx_OutgoingFlows)` ?

cpp/models/ide_secir/model.cpp
```diff
@@ -195,49 +250,93 @@ void Model::compute_flow(int idx_InfectionTransitions, Eigen::Index idx_Incoming
...
     compute_flow((int)InfectionTransition::InfectedCriticalToDead,
                  Eigen::Index(InfectionTransition::InfectedSevereToInfectedCritical), dt);
-    // Flow from U to R
+    // Flow InfectedCriticalToRecovered.
     compute_flow((int)InfectionTransition::InfectedCriticalToRecovered,
                  Eigen::Index(InfectionTransition::InfectedSevereToInfectedCritical), dt);
 }
 
-void Model::compute_deaths()
+void Model::update_compartments_current_timestep()
```
mknaranja, line 287: **Important** Why is susceptible not reduced here???

mknaranja (changes requested):
Thank you @lenaploetzke for these changes. I mostly have small suggestions (which however might change code lines in many occasions). Could you check them?

cpp/models/ide_secir/model.h
```diff
@@ -153,12 +139,13 @@ class Model
     void flows_current_timestep(ScalarType dt);
 
     /**
-     * @brief Computes total number of people in the #Dead compartment for the current last time in m_populations.
-     * 
-     * Number is stored in m_populations.
+     * @brief Sets the values of compartments (all apart from #Susceptible) for the current last timestep in m_populations.
      *
+     * New values are stored in m_populations. The values are calculated using the compartment size in the previous time step and the related flows of the current time step. 
+     * Therefore the flows of the current time step should be calculated before using this function.
```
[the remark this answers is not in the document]
lenaploetzke: I actually had this setting but ignored it :)

cpp/models/ide_secir/model.cpp
lenaploetzke: This is because we need the Susceptibles of the current time step to calculate the SusceptibleToExposed transition. Therefore, S must be calculated before the transitions. To update the other compartments, we need the transitions of the current time step, so the compartments have to be calculated after the transitions. This means that the calculation order in the simulation function cannot be changed and S cannot be calculated with the other compartments!

cpp/models/ide_secir/model.cpp
lenaploetzke: In principle, we could simply implement it, but then we would have one more function, making it more complex. Would you like to have such a function?

mknaranja: Not sure, just thought about simpler extension to new models (maybe make it less error prone with the +/- signs?) 
Maybe you can have a 5min discussion with @reneSchm and then you decide ;)

cpp/models/ide_secir/infection_state.h
```diff
@@ -61,7 +61,8 @@ enum class InfectionTransition
     InfectedSevereToRecovered            = 7,
     InfectedCriticalToDead               = 8,
     InfectedCriticalToRecovered          = 9,
-    Count                                = 10
+    Count                                = 10,
+    DummyIndex                           = 0
```
mknaranja, line 65: Is this okay to use 0 twice?

mknaranja (approved):

mknaranja (approved):

PR #761, opened 2023-09-04
636 make python serialization usable again

### LINKED ISSUE #636: Make Python serialization usable again
### Feature description

The python package memilio-simulation already implements methods for binding classes that support serilization in c++ and a serilizer. Nevertheless serilization is not usable for most classes in Python. This is due to the fact, that in order to activate it in Python the method of binding need to be changed to the one for serializable classes.
To make it better usable we want to make the decision if a class can be transfered to python with serialization or not at compile time. Thus in future all classes need to be declared for python through a pymio method instead, which internally deduces the way of binding, instead of using either the pybind11 method or pymio method for serialisable classes. 
As a result all future classes are automatically serilisable in python, if that is the case in c++, without needing to change any of the binding code.

### Additional context

_No response_

### Checklist

- [X] Linked to project.

MaxBetzDLR:
# Changes and Information

-

### Checks by code author

Closes #636


pycode/memilio-simulation/memilio/simulation/pybind_util.h
```diff
@@ -121,15 +193,18 @@ auto bind_Range(pybind11::module_& m, const std::string& class_name)
             pybind11::return_value_policy::reference_internal);
 
     //bindings for the range itself
-    pybind11::class_<Range>(m, class_name.c_str())
+    bind_class<PickleFlag::NoPickling, Range>(m, class_name.c_str())
```
dabele, line 196: When would you choose which PickleFlag? Would it be possible to document somwhere a simple rule of thumb?

pycode/memilio-simulation/memilio/simulation/pybind_util.h
```diff
@@ -52,7 +96,34 @@ decltype(auto) pybind_pickle_class(pybind11::module_& m, const char* name)
...
+
+template <class T, class... Args>
+decltype(auto) _bind_class(pybind11::module& m, std::string const& name, ForcePicklingTag) {
+    decltype(auto) cls = pybind11::class_<T, Args...>(m, name.c_str());
+    pybind_pickle_class<T, Args...>(cls);
+    return cls;
+}
+
+template <PickleFlag F, class T, class... Args>
+decltype(auto) bind_class(pybind11::module& m, std::string const& name) {
```
dabele, line 125: what's the purpose of decltype(auto) here? Does pybind11::class_ ever return a reference?

pycode/memilio-simulation/memilio/simulation/abm.cpp
```diff
@@ -69,7 +69,7 @@ PYBIND11_MODULE(_simulation_abm, m)
         .value("PublicTransport", mio::abm::LocationType::PublicTransport)
         .value("TransportWithoutContact", mio::abm::LocationType::TransportWithoutContact);
 
-    py::class_<mio::abm::TestParameters>(m, "TestParameters")
+    pymio::bind_class<pymio::PickleFlag::TryPickling, mio::abm::TestParameters>(m, "TestParameters")
```
dabele, line 72: I think types like this should be ForcePickling. Otherwise someone adds a parameter that isn't serializable and the pickling disappears. I think TryPickling only makes sense for generic types.

pycode/memilio-simulation/memilio/simulation/abm.cpp
```diff
@@ -82,7 +82,7 @@ PYBIND11_MODULE(_simulation_abm, m)
     pymio::bind_ParameterSet<mio::abm::LocalInfectionParameters>(m, "LocalInfectionParameters").def(py::init<>());
     pymio::bind_ParameterSet<mio::abm::MigrationParameters>(m, "MigrationParameters").def(py::init<>());
 
-    py::class_<mio::abm::TimeSpan>(m, "TimeSpan")
+    pymio::bind_class<pymio::PickleFlag::TryPickling, mio::abm::TimeSpan>(m, "TimeSpan")
```
dabele, line 85: why wouldn't this be ForcePickling?

dabele (changes requested):
I like the general design, but have a few concerns about the implementation. I also don't understand some of the choices of TryPickling/ForcePickling/NoPickling, but I only added comments for some cases. Maybe try to explain your general reasoning there.

pycode/memilio-simulation/memilio/simulation/pybind_util.h
```diff
@@ -22,18 +22,62 @@
...
 #include "pybind11/pybind11.h"
 
 namespace pymio
 {
 
-//bind class and add pickling based on memilio serialization framework
+enum class PickleFlag {
+    NoPickling,
+    TryPickling,
+    ForcePickling
```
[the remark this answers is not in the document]
MaxBetzDLR: That are some good suggestions

pycode/memilio-simulation/memilio/simulation/pybind_util.h
```diff
@@ -52,7 +96,34 @@ decltype(auto) pybind_pickle_class(pybind11::module_& m, const char* name)
...
+decltype(auto) _bind_class(pybind11::module& m, std::string const& name, ForcePicklingTag) {
+    decltype(auto) cls = pybind11::class_<T, Args...>(m, name.c_str());
+    pybind_pickle_class<T, Args...>(cls);
+    return cls;
+}
+
+template <PickleFlag F, class T, class... Args>
+decltype(auto) bind_class(pybind11::module& m, std::string const& name) {
+    return _bind_class<T, Args...>(m, name, typename PicklingTag<F>::type{});
 }
```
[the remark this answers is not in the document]
MaxBetzDLR: My thinking was to not divide the main class T and base classes Args. But i see that the class T is the most important part.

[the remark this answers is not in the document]
dabele: Good point. I would still prefer pickle<Derived, PicklingTag, Base> over the alternative. I think that's readable enough. 

I guess the most generic version would be to make the PicklingTag a part of `Args...` (can't be an enum then) and iterate over Args at compile time to find it. Then it could be anywhere, and other flags could be added in the same way when needed. But I think that's not worth the effort now.

pycode/memilio-simulation/memilio/simulation/pybind_util.h
MaxBetzDLR: it was used similar in other parts of the code for pybind11::class_, but i dont think it will ever return a reference.

pycode/memilio-simulation/memilio/simulation/abm.cpp
MaxBetzDLR: It probably should be in the future, but for now that results in a compilation error, because TimeSpan is not serializable.

dabele (changes requested):

dabele (approved):

PR #770, opened 2023-09-15
769 new initialization for ide model

### LINKED ISSUE #769: New Initialization for IDE model

With the current initialization method, the susceptibles are overestimated under certain parameter choices, so that S(0)>N_0.

### Enhancement description

Different initialization methode adapted to the available data from the RKI.

### Additional context

_No response_

### Checklist

- [X] Linked to project

lenaploetzke:
# Changes and Information

- added a different method to initialize the ide model. Is comparable to the method used for the ODE model for compartment Recovered. Examples and tests run correctly.

### Checks by code author

Closes #769


cpp/tests/test_ide_secir.cpp
```diff
@@ -242,15 +242,181 @@ TEST(IdeSecir, checkSimulationFunctions)
...
+    sim4.advance(tmax);
+    result = sim4.get_result();
+
+    // Compare with previous run.
+    for (Eigen::Index i = 0; i < (Eigen::Index)mio::isecir::InfectionState::Count; i++) {
+        EXPECT_NEAR(result[0][i], vec_R[i], 1e-8);
+    }
+
+    // --- Case without fitting initialization method.
+    mio::isecir::Model model5(std::move(init), N, deaths, 0, std::move(parameters));
```
annawendler, line 393: I would make this test more explicit and set init to the zero vector. If you use move twice you don't know what you are putting in your model. In theory it could be possible that this is a valid init vector for the forceofinfection method.

cpp/models/ide_secir/model.cpp
```diff
@@ -29,64 +29,90 @@ namespace mio
...
 
-Model::Model(TimeSeries<ScalarType>&& init, ScalarType N_init, ScalarType Dead_before,
+Model::Model(TimeSeries<ScalarType>&& init, ScalarType N_init, ScalarType deaths, ScalarType total_confirmed_cases,
              const ParameterSet& Parameterset_init)
     : parameters{Parameterset_init}
     , m_transitions{std::move(init)}
     , m_populations{TimeSeries<ScalarType>(Eigen::Index(InfectionState::Count))}
     , m_N{N_init}
-    , m_deaths_before{Dead_before}
+    , m_total_confirmed_cases(total_confirmed_cases)
```
annawendler, line 38: for consistency
```suggestion
    , m_total_confirmed_cases{total_confirmed_cases}
```

cpp/models/ide_secir/model.cpp
```diff
@@ -29,64 +29,90 @@ namespace mio
...
+                m_populations[Eigen::Index(0)][Eigen::Index(InfectionState::InfectedSymptoms)] -
+                m_populations[Eigen::Index(0)][Eigen::Index(InfectionState::InfectedSevere)] -
+                m_populations[Eigen::Index(0)][Eigen::Index(InfectionState::InfectedCritical)] -
+                m_populations[Eigen::Index(0)][Eigen::Index(InfectionState::Recovered)] -
+                m_populations[Eigen::Index(0)][Eigen::Index(InfectionState::Dead)];
+        }
+        else {
+            log_error("Error occured while initializing compartments: Force of infection is evaluated to 0 and neither "
+                      "Susceptibles nor Recovered for time 0 were set. One of them should be larger 0.");
+        }
```
annawendler, line 105: I think the error message should also include total_confirmed_cases.
Also, I would find it more intuitive if the distinction between `m_total_confirmed_cases < 1e-12` and the other case was the other way around, i.e. if it would be checked if `m_total_confirmed_cases > 1e-12` first.

cpp/models/ide_secir/model.h
```diff
@@ -195,6 +196,16 @@ class Model
...
+    /**
+     * @brief Setter for total number of confirmed cases at time t0.
+     * 
+     * @param[in] total_confirmed_cases Total confirmed cases at time t0 can be set if it should be used for initialisation.
+     */
+    void set_total_confirmed_cases(ScalarType total_confirmed_cases)
+    {
+        m_total_confirmed_cases = total_confirmed_cases;
+    }
+
```
annawendler, line 208: What is the use case for this function? I would think that we either know total_confirmed_cases at the beginning of the simulation or not. If we know it, we can include it directly in the construction of the model and don't need to set this value at a later point but I may be missing something.

cpp/models/ide_secir/model.h
```diff
@@ -205,7 +216,8 @@ class Model
 private:
     ScalarType m_forceofinfection{0}; ///< Force of infection term needed for numerical scheme.
     ScalarType m_N{0}; ///< Total population size of the considered region.
-    ScalarType m_deaths_before{0}; ///< Deaths before start of simulation (at time -m_dt).
+    ScalarType m_deaths_before{0}; ///< Total number of deaths at the time point - dt_init.
```
annawendler, line 219: Is dt_init from the LCT implementation?

cpp/tests/test_ide_secir.cpp
```diff
@@ -242,15 +242,181 @@ TEST(IdeSecir, checkSimulationFunctions)
     }
 }
 
+// Check if the model uses uses the correct method for initialization.
```
annawendler, line 245:
```suggestion
// Check if the model uses the correct method for initialization.
```

annawendler (changes requested):
Nice initialization! I added a few comments and questions, especially regarding the test.

cpp/tests/test_ide_secir.cpp
lenaploetzke: the contact rate was previously set to 0. Therefore the vector can be valid here. If the force infection method worked, it would also be used in cases S and R. Should I still make this more explicit and pass the zero vector? 
That's the first and last time i use move with the vector, so this is ok, isn't it?

cpp/models/ide_secir/model.h
lenaploetzke: Yes, this setter is not used, i deleted it

cpp/tests/test_ide_secir.cpp
```diff
@@ -242,15 +242,181 @@ TEST(IdeSecir, checkSimulationFunctions)
...
+// This is done by comparing the compartment sizes at t0 with a previous run with the same initialization method.
```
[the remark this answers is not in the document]
lenaploetzke: Idea was to target all the method for initialization. There is no way to get which method was used to initialize, so I simply compared the results to a previous run. The idea is that the same method was used when we have same results. If the results don't match, something has gone wrong. I created the results with a simulation in which I checked (via debug mode) which method was used. Are you okay with this idea or maybe a better suggestion?

cpp/tests/test_ide_secir.cpp
```diff
@@ -242,15 +242,181 @@ TEST(IdeSecir, checkSimulationFunctions)
...
+    result = sim2.get_result();
+
+    // Compare with previous run.
+    for (Eigen::Index i = 0; i < (Eigen::Index)mio::isecir::InfectionState::Count; i++) {
+        EXPECT_NEAR(result[0][i], vec_total_confirmed[i], 1e-8);
+    }
+
+    // --- Case with S.
+    contact_matrix[0]                              = mio::ContactMatrix(Eigen::MatrixXd::Constant(1, 1, 0));
+    parameters.get<mio::isecir::ContactPatterns>() = mio::UncertainContactMatrix(contact_matrix);
```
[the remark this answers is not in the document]
lenaploetzke: Yes. Added a comment

cpp/tests/test_ide_secir.cpp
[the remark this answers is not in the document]
annawendler: Thanks for the explanation. The issue I see here is that the test depends on the results of the respective initialization method although what is tested is not if these results are correct but that the correct if/else statement was selected. This way the test needs to be maintained and changed every time an initialization method is changed (although I don't see how this would change in the near future).
One idea to avoid this would be to set m_total_confirmed_cases =-1,-2 etc. (or just some value that is not valid for the initialization method using m_total_confirmed_cases) depending on the initialization option and to check the value of m_total_confirmed_cases in the test.

cpp/tests/test_ide_secir.cpp
annawendler: What do you mean with the sentence "If the force infection method worked, it would also be used in cases S and R."?

You're right, if the contact rate is set to 0 it doesn't matter what vector is used for init because the forceofinfection will always be 0. Then it's fine if you use move, I think a comment would still be helpful here and make it more obvious :)

lenaploetzke: > What do you mean with the sentence "If the force infection method worked, it would also be used in cases S and R."?
>

If the forceofinfection was greater than zero, this method would also be used to initialize in the tests for the initialization with S and R (see lines 353-395).  I think the added comment in line 354f. makes this more clear :) 

> You're right, if the contact rate is set to 0 it doesn't matter what vector is used for init because the forceofinfection will always be 0. Then it's fine if you use move, I think a comment would still be helpful here and make it more obvious :)

Added a comment.

cpp/tests/test_ide_secir.cpp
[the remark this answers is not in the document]
lenaploetzke: I rewrote the test and added a function to the class to get the method for initialization.

annawendler (approved):

PR #795, opened 2023-10-09
First draft.

### LINKED ISSUE #501: Add Intervention Scheme for ABM
With #257 we implemented mask usage. Furthermore, we can prevent people from going to certain locations, however, we do not yet have a universal and clear `InterventionScheme`.

The problem with the current implementation of Masks is:

1. You cannot require "no mask" as MaskType = None is not implemented.
2. If MaskType = None was implemented, then "time_used" would be reset and everything I would reuse my previous mask, time would start from zero. This could be "solved" by having two mask_types `m_current` and `m_preference`. The current could be None and time_used would point to preference if no mask is worn at the moment. But this is not very elegant. 
3. Another problem is that for each contact reduction intervention, like `reduced maximum capacity`, `minimal_distance` or so, we would need at least one other member variable.

From 1.-3. I came to the idea that we would maybe template the location as

```
template<typename... LocalInterventions>
class Location
{
    // ...
};
```

and then I could maybe have

`Location<Masks>` or `Location<Masks, MinimalDistance>`

where only the considered interventions are in the list. When designing this step by step it just came to my head that we cannot only use the active NPIs in the list since this will be set a compile time. So we would then maybe even need start and end dates for the interventions...

However, I have no good design / idea on the remaining part of this proposed change.

Maybe we also go with an `InterventionScheme` as the `TestingScheme` which is on Locations and ...

_Originally posted by @mknaranja in https://github.com/DLR-SC/memilio/pull/471#discussion_r1031376772_

DavidKerkmann:
# Changes and Information

-

### Checks by code author

- [ ] New code adheres to [coding guidelines](https://github.com/DLR-SC/memilio/wiki/Coding-guidelines)

- [ ] Appropriate **unit tests** have been added, CI passes and code coverage is acceptable (did not decrease)

closes #501

### DIFF
--- a/cpp/models/abm/npi.cpp
+++ b/cpp/models/abm/npi.cpp
@@ -1,0 +1,149 @@
+/* 
+* Copyright (C) 2020-2023 German Aerospace Center (DLR-SC)
+*
+* Authors: David Kerkman
+*
+* Contact: Martin J. Kuehn <Martin.Kuehn@DLR.de>
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#include "abm/npi.h"
+#include "abm/person.h"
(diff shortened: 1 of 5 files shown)


cpp/models/abm/npi.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "abm/npi.h"
+#include "abm/person.h"
+#include "abm/location.h"
+
+namespace mio
+{
+namespace abm
+{
+
+struct AcceptBaseStringency : AcceptBase {
```
mknaranja, line 30: Could you provide some comments on this struct on how / what it is meant to be (used)?

DavidKerkmann: I have done so.

PR #830, opened 2023-11-11
829 optimize testingstrategy

### LINKED ISSUE #357: Test container for TestingCriteria
For performance issues, we should test which data container should be used in TestingCriteria. For now, a vector is used, but a set or an unordered set might be better when having many lookups. This should be tested and changed, if another container is faster.

### LINKED ISSUE #829: Further optimization of TestingStrategy

Beyond the optimization in #752, `TestingStrategy` is still very expensive. Worse, it is very expensive even if no TestingSchemes are added at all. 

### Enhancement description

Performance optimizations of `TestingStrategy`

### Additional context

commenting out the testing in `World::migration` reduces runtime by ~30-50%

### Checklist

- [X] Linked to project

### LINKED ISSUE #857: segfault in ABM with OpenMP
### Bug description

OpenMP multithreaded abm_simulation crashes with a segfault. without OpenMP (or OMP_NUM_THREADS=1) the program runs normal. In Release build the crash happened reliably for me. In Debug builds the crash doesn't always happen or takes a long time to happen, so it's hard to debug

### Version

Linux

### To reproduce

1. build abm_simulation with MEMILIO_ENABLE_OPENMP=ON
2. `export OMP_NUM_THREADS=<N>`, higher N _seems_ to increase chance of it happening.
3. run `abm_simulation 1 <dir>`

### Relevant log output

_No response_

```zsh
gcc 11, ubuntu 20
```

### Checklist

- [X] Linked to project

dabele:
## Changes and Information

@khoanguyen-dev @xsaschako @DavidKerkmann 
These changes are not ready for review yet, at least some need to be discussed since they change the model, not just performance. I want to show the changes early as a basis for discussion.

Closes #829 
Closes #357 
Closes #857

### 1
use bitset for age groups in `TestingCriteria`; bitset needs a fixed size, so add a maximum number of age groups. performance gain is quite small (< 1-2%), so maybe not necessary if you feel the maximum number is not wanted.

### 2
switch the order of `if` for probabilty check and evaluating `TestingCriteria` when applying the test; again a very small performance gain (<5%) that also depends on the performance of the RNG and the complexity of `TestingCriteria`. if TestingCriteria might become more complex in the future we may want to skip this. This doesn't change the model in theory, but in practice it does since the sequence of random numbers changes.

### 3
use CustomIndexArray for AgeGroupGoToSchool/Work; the performance gain here is barely measureable. But the interface of CustomIndexArray is more consistent with the rest of the parameters; AgeGroup and the enums we use are valid indices for a reason, random access is fast and convenient using an index into an array.

### 4
use std::vector to store testingschemes per location instead of std::map; The performance gains here are very large (~20-30% of total benchmark runtime). The current implementation with map has two problems: 
    - lookups in map are more expensive than linear search in vectors for small numbers of elements; map is only better for quite large numbers. 
    - lookup in map with `operator[]` adds an element to the map if it doesn't exist. So the map quickly grows and has an entry for every location, making lookup even slower. This can be solved by using std::vector, but it can also be solved with a map by using `map::find` instead of `map::operator[]`, so it's at least somewhat independent from the first issue, but the implementation with vector is slightly faster, at least unless you add a lot of TestingSchemes for different locations

### 5
only run the TestingStrategy when migrating; this is the largest change to the model, but also the largest performance gain, ~30-50% (Note that these are not fully additive with the other improvements above, if the TestingStrategy is executed less often, the runtime of the TestingStrategy itself doesn't matter as much); Persons are tested much less, persons that never leave home are never tested at all, even if there is a TestingScheme for home locations; that may be realistic, but needs to be discussed. 

Another model could be to run tests only when migrating, but for the source location as well as the destination. This would give most of the performance benefits from this change with probably less impact on the model results (e.g. person leaves home to get to work, is tested at home and at work and stays home if either of them fails). 

If we expect a lot of TestingSchemes for specific single locations (instead of locationtypes) it might be better to store the testing scheme in the location directly.

### Checks by code author

### DIFF
--- a/cpp/models/abm/parameters.h
+++ b/cpp/models/abm/parameters.h
@@ -497,10 +497,12 @@ struct GotoSchoolTimeMaximum {
  * @brief The set of AgeGroups that can go to school.
  */
 struct AgeGroupGotoSchool {
-    using Type = std::set<AgeGroup>;
-    static Type get_default(AgeGroup /*size*/)
+    using Type = CustomIndexArray<bool, AgeGroup>;
+    static Type get_default(AgeGroup num_agegroups)
     {
-        return std::set<AgeGroup>{AgeGroup(1)};
+        auto a = Type(num_agegroups, false);
+        a[AgeGroup(1)] = true;
+        return a;
     }
     static std::string name()
     {
@@ -512,10 +514,13 @@ struct AgeGroupGotoSchool {
  * @brief The set of AgeGroups that can go to work.
  */
 struct AgeGroupGotoWork {
-    using Type = std::set<AgeGroup>;
-    static Type get_default(AgeGroup /*size*/)
+    using Type = CustomIndexArray<bool, AgeGroup>;
+    static Type get_default(AgeGroup num_agegroups)
     {
-        return std::set<AgeGroup>{AgeGroup(2), AgeGroup(3)};
+        auto a = Type(num_agegroups, false);
+        a[AgeGroup(2)] = true;
+        a[AgeGroup(3)] = true;
+        return a;
     }
     static std::string name()
     {
--- a/cpp/examples/abm_minimal.cpp
+++ b/cpp/examples/abm_minimal.cpp
@@ -41,9 +41,12 @@ int main()
     world.parameters.get<mio::abm::IncubationPeriod>() = 4.;
 
     // Set the age group the can go to school is AgeGroup(1) (i.e. 5-14)
-    world.parameters.get<mio::abm::AgeGroupGotoSchool>() = {age_group_5_to_14};
+    world.parameters.get<mio::abm::AgeGroupGotoSchool>() = false;
+    world.parameters.get<mio::abm::AgeGroupGotoSchool>()[age_group_5_to_14] = true;
     // Set the age group the can go to work is AgeGroup(2) and AgeGroup(3) (i.e. 15-34 and 35-59)
-    world.parameters.get<mio::abm::AgeGroupGotoWork>() = {age_group_15_to_34, age_group_35_to_59};
+    world.parameters.get<mio::abm::AgeGroupGotoWork>() = false;
+    world.parameters.get<mio::abm::AgeGroupGotoWork>()[age_group_15_to_34] = true;
+    world.parameters.get<mio::abm::AgeGroupGotoWork>()[age_group_35_to_59] = true;
 
     // Check if the parameters satisfy their contraints.
     world.parameters.check_constraints();
--- a/cpp/models/abm/config.h
+++ b/cpp/models/abm/config.h
@@ -1,0 +1,36 @@
+/*
+* Copyright (C) 2020-2024 MEmilio
+*
+* Authors: Daniel Abele
+*
+* Contact: Martin J. Kuehn <Martin.Kuehn@DLR.de>
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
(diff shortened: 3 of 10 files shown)


dabele:
A few general notes
- map and set may be semantically convient in many cases, but are generally slow. unordered_set and map are much faster, but still slower than std::vector in many cases. these containers are generally only worth it for large data sets with sparse random accesses. Cutoff points need to be measured of course, but vector/array/CustomIndexArray is usually a good default choice, especially when the the data (AgeGroup, enums) can be used as an index. 
- map and unordered_map are also dangerous because `operator[]` adds an entry if it doesn't find it
- order of conditions matters
- we need to find ways to not do everything in every iteration/time step whenever possible. Even checking a single bool can be slow if it isn't in the CPU cache i.e. if it isn't used regularly.

DavidKerkmann:
This together with #752 will also close #357.

dabele:
Cleaned up the changes and fixed the unit tests.

Did better benchmark measurements to asses the changes. Uncertainty of ~2%.

baseline (current main): 8006 ms
with change 4 (vector instead of unordered_map for testing schemes): 6088 ms (~25%)
with change 4 and 2 (switch of ifs in testing strategy): 5640 ms (additional ~8 %)
with change 4, 2 and 3 (CustomIndexArray for GoToWork/GoToSchool paramater): 5638 ms (no measureable change)
with change 4, 2, 3 and 1 (bitset for agegroups in TestingCriteria with fixed number of age groups): 5262 ms (additional ~7%)

From our discussions changes 2,3,4 should be fine to add. Change 3 does not give performance, just cleaner code (I think). Change 1 adds some complexity to the model, we can discuss whether the performance gain is worth it. 

Reverted change 5 (only run TestingStrategy when migrating) as discussed. But I measured the performance for this change as well. Only testing persons that migrate to a different location gives 3889 ms (~27%). This measurement includes all the other changes above, I did not measure the change independently. So we probably should keep this in the discussion but in a separate Issue/MR.

cpp/models/abm/parameters.h
```diff
@@ -512,10 +514,13 @@ struct AgeGroupGotoSchool {
...
 struct AgeGroupGotoWork {
-    using Type = std::set<AgeGroup>;
-    static Type get_default(AgeGroup /*size*/)
+    using Type = CustomIndexArray<bool, AgeGroup>;
+    static Type get_default(AgeGroup num_agegroups)
     {
-        return std::set<AgeGroup>{AgeGroup(2), AgeGroup(3)};
+        auto a = Type(num_agegroups, false);
+        a[AgeGroup(2)] = true;
+        a[AgeGroup(3)] = true;
```
DavidKerkmann, line 522: This could be a bit problematic if the number of age groups is lower than 3. This was already a problem earlier I guess, but could be worse now due to the indexing. Perhaps we could return a more standard default case and rather set this parameter in each simulation?

DavidKerkmann (commented):
Thanks for the improvements. I agree that we should track point 5 for later inclusion as this greatly increases performance and is important. For now, I only have one remark.

cpp/models/abm/parameters.h
dabele: I am worried about this as well. The usage is a bit awkward as well, first setting the whole array to false and some of them to true (see the abm unit tests). what would the default be? all false?

DavidKerkmann: I guess all false makes sense, and then set up an initialisation script that sets parameters (also others potentially) to values that we commonly use in tests/simulations. This is beyond this PR though, so I am inclined to leave it as it is and open a new issue for setting up parameters separately. This is a step that we wanted to to anyway at some point.

dabele: created issue #869

DavidKerkmann (approved):

mknaranja (approved):

PR #863, opened 2023-12-07
757 dont store pointers in the abm

reneSchm:
# Changes and Information

- remove pointers from World, Location, Person
- this may change quite a bit more

### Checks by code author

- [x] (For ABM development) Checked [benchmark results](https://github.com/SciCompMod/memilio/wiki/Agent-Based-Model-Development)

```
abm_benchmark/abm_benchmark_50k        4290 ms         4288 ms            1
abm_benchmark/abm_benchmark_100k       9741 ms         9737 ms            1
abm_benchmark/abm_benchmark_200k      20698 ms        20691 ms            1
```


reneSchm:
I have a few open items that may need a bit of discussion:
- Naming: I moved several methods around, without changing their names. None of them are particularly bad, but I think there should be better names out there - I would really appreciate some suggestions. Some examples:
  - `get_number_persons` returns #persons at a location, which in general is not the same as the size of `get_persons`
  - `get_subpopulation` is okay, but does not fit well with `get_number_persons`
  - The names `migrate` and `interact` both in the World and as free functions are not really descriptive
  - The header `functions.h` still needs a more meaningful name. It contains ABM defining free functions (i.e. non class member functions) like migrate and interact.
- Caching: Currently, I only cache the population size at each location. As far as I can tell that is the only metric used during simulations, but we are missing e.g. cell populations or persons of a given infection state. The latter would be used by get_subpopulations. Should I stratify the cache or add more caches? If so, which?
- Should we add functions like Person::migrate(_to) back? I don't think so, since we only replace one argument by the calling class, and it adds more code to maintain without adding any functionality. 
- For now I kept most tests running, but I think they need a bigger rework. I suggest doing that in another issue, since I barely changed any effective behavior here (but a lot of internals/interfaces). This way we could merge this PR faster.
- I marked the move (assignment) constructor for World as default, so they will just make a copy using the custom copy ctor. I don't think we gain much here by actually moving something, but we could define a custom move ctor anyways.

reneSchm:
On performance: There are some small improvements, but unsurprisingly nothing big. Here is single threaded (OMP_NUM_THREADS=1) performance of the current main branch as reference:
```
Running ../build/bin/abm_benchmark
Run on (56 X 3700 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x56)
  L1 Instruction 32 KiB (x56)
  L2 Unified 1024 KiB (x56)
  L3 Unified 19712 KiB (x4)
Load Average: 1.03, 2.78, 1.36
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
---------------------------------------------------------------------------
Benchmark                                 Time             CPU   Iterations
---------------------------------------------------------------------------
abm_benchmark/abm_benchmark_50k        4617 ms         4615 ms            1
abm_benchmark/abm_benchmark_100k      10602 ms        10598 ms            1
abm_benchmark/abm_benchmark_200k      23102 ms        23094 ms            1
```
All benchmarks are made on the HPDA Cluster using one node (with --exclusive). I will leave out the preamble now, since it does not change (apart from the load average, which is usually less then 2.00).

Here are the current results for this PR:
```
abm_benchmark/abm_benchmark_50k        4290 ms         4288 ms            1
abm_benchmark/abm_benchmark_100k       9741 ms         9737 ms            1
abm_benchmark/abm_benchmark_200k      20698 ms        20691 ms            1
```
So we got very roughly a 5-10% performance gain.

Now to the parallel performance. This is more difficult to measure, as the timing results seem somewhat random. Perhaps this is due to the CPU scaling, but at least it is consistent in both branches. Here is an example for 20 threads, launched with
`(OMP_NUM_THREADS=20 ./build/bin/abm_benchmark) && (OMP_NUM_THREADS=20 ./build/bin/abm_benchmark)`, i.e. the second benchmark is run immediately after the first:
```
abm_benchmark/abm_benchmark_50k        2078 ms         2077 ms            1
abm_benchmark/abm_benchmark_100k       4364 ms         4363 ms            1
abm_benchmark/abm_benchmark_200k       9036 ms         9033 ms            1
---------------------------------------------------------------------------
abm_benchmark/abm_benchmark_50k        1115 ms         1115 ms            1
abm_benchmark/abm_benchmark_100k       2354 ms         2353 ms            1
abm_benchmark/abm_benchmark_200k       5316 ms         5312 ms            1
```
Note that the scaling seems to be consistent with the first run's time. But that time changes randomly every time. Also, the second run may be the slower of the two, or equally as fast. I see no pattern here.

Anyways, to get just a quick idea how well this branch scales, I ran 10 benchmarks and picked *only the fastest* result:
```
--1-thread-----------------------------------------------------------------
abm_benchmark/abm_benchmark_50k        4267 ms         4264 ms            1
abm_benchmark/abm_benchmark_100k       9820 ms         9816 ms            1
abm_benchmark/abm_benchmark_200k      20881 ms        20874 ms            1
--2-threads----------------------------------------------------------------
abm_benchmark/abm_benchmark_50k        2368 ms         2366 ms            1
abm_benchmark/abm_benchmark_100k       5415 ms         5412 ms            1
abm_benchmark/abm_benchmark_200k      11600 ms        11596 ms            1
--4-threads----------------------------------------------------------------
abm_benchmark/abm_benchmark_50k        1676 ms         1675 ms            1
abm_benchmark/abm_benchmark_100k       3839 ms         3837 ms            1
abm_benchmark/abm_benchmark_200k       8238 ms         8232 ms            1
--8-threads----------------------------------------------------------------
abm_benchmark/abm_benchmark_50k        1364 ms         1363 ms            1
abm_benchmark/abm_benchmark_100k       2914 ms         2913 ms            1
abm_benchmark/abm_benchmark_200k       6412 ms         6391 ms            1
-16-threads----------------------------------------------------------------
abm_benchmark/abm_benchmark_50k         931 ms          930 ms            1
abm_benchmark/abm_benchmark_100k       1918 ms         1917 ms            1
abm_benchmark/abm_benchmark_200k       4798 ms         4791 ms            1
-32-threads----------------------------------------------------------------
abm_benchmark/abm_benchmark_50k        1068 ms         1068 ms            1
abm_benchmark/abm_benchmark_100k       2264 ms         2263 ms            1
abm_benchmark/abm_benchmark_200k       5218 ms         5214 ms            1
```

cpp/models/abm/personal_rng.cpp
```diff
@@ -0,0 +1,42 @@
+/* 
+* Copyright (C) 2020-2024 MEmilio
+*
+* Authors: Daniel Abele, Elisabeth Kluth, David Kerkmann, Khoa Nguyen, Rene Schmieding
```
khoanguyen-dev, line 4:
```suggestion
* Authors: Rene Schmieding
```

cpp/models/abm/personal_rng.h
```diff
@@ -0,0 +1,100 @@
+/* 
+* Copyright (C) 2020-2024 MEmilio
+*
+* Authors: Daniel Abele, Elisabeth Kluth, David Kerkmann, Khoa Nguyen, Rene Schmieding
```
khoanguyen-dev, line 4:
```suggestion
* Authors: Rene Schmieding
```

cpp/models/abm/testing_strategy.h
```diff
@@ -136,12 +136,12 @@ class TestingScheme
 
     /**
      * @brief Runs the TestingScheme and potentially tests a Person.
-     * @param[inout] rng Person::RandomNumberGenerator for the Person being tested.
+     * @param[inout] rng PersonalRandomNumberGenerator for the Person being tested.
```
khoanguyen-dev, line 139:
```suggestion
     * @param[inout] rng PersonalRandomNumberGenerator of the Person being tested.
```

cpp/models/abm/testing_strategy.h
```diff
@@ -213,13 +213,13 @@ class TestingStrategy
 
     /**
      * @brief Runs the TestingStrategy and potentially tests a Person.
-     * @param[inout] rng Person::RandomNumberGenerator for the Person being tested.
+     * @param[inout] rng PersonalRandomNumberGenerator for the Person being tested.
```
khoanguyen-dev, line 216:
```suggestion
     * @param[inout] rng PersonalRandomNumberGenerator of the Person being tested.
```

cpp/models/abm/location_type.h
```diff
@@ -76,6 +77,48 @@ struct LocationId {
...
+    uint32_t get_index() const
+    {
+        return index;
+    }
+
+    // same interface as location
+    LocationType get_type() const
+    {
+        return type;
+    }
```
khoanguyen-dev, line 91: Do you need these two functions? Both members of LocationId are publicly accessible, right?

reneSchm: No, they are not necessary. They do allow using the same interface for Location and its ID to access type/index, which saved me some additional work from changing several interfaces from using Location to LocationId. I could argue that getter/setters can be better than direct access, but these functions are here mostly for convenience.

cpp/simulations/abm_braunschweig.cpp
```diff
@@ -364,12 +366,13 @@ void create_world_from_data(mio::abm::World& world, const std::string& filename,
...
-            auto& person           = world.add_person(first_location, determine_age_group(age));
-            auto home              = locations.find(home_id)->second;
+            auto& person =
+                world.get_person(world.add_person(first_location, determine_age_group(age))); // TODO: is this safe?
+            auto home = locations.find(home_id)->second;
             person.set_assigned_location(home);
             person.set_assigned_location(hospital);
             person.set_assigned_location(icu);
-            persons.insert({person_id, person});
+            persons.insert({person_id, person}); // TODO: why? what?
```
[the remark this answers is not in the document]
reneSchm: There are three, to be exact.

But granted, they are not especially well formulated. I think I was confused what this code is doing, in particular why we are adding a person twice (first to world, then using the insert, which is a reference to the world population). That is also why I ask if the line above is safe, because insertion here invalidates the Person& (which was not an issue with Person*)

cpp/tests/test_abm_world.cpp
```diff
@@ -165,7 +170,7 @@ TEST(TestWorld, evolveStateTransition)
     EXPECT_EQ(p3.get_infection_state(t + dt), mio::abm::InfectionState::InfectedSymptoms);
 }
 
-TEST(TestWorld, evolveMigration)
+TEST(TestWorld, evolveMigration) // TODO: this is not a unit test, this is a unit of a test!
```
DavidKerkmann, line 173: Yes, it basically tests certain aspects of the evolve function, which is basically all other big functions together. Do you have ideas how to make a simpler unit test for this?

cpp/tests/test_abm_world.cpp
```diff
@@ -545,179 +558,3 @@ TEST(TestWorld, checkParameterConstraints)
     params.get<mio::abm::LockdownDate>() = mio::abm::TimePoint(-2);
     ASSERT_EQ(params.check_constraints(), true);
 }
-
-TEST(TestWorld, copyWorld)
```
DavidKerkmann, line 549: Why was this removed? Because we use default (or almost default) copy constructor?

DavidKerkmann (commented):
Mostly small requests/suggestions. I tried to answer some of the open questions and asked some new questions. Maybe we can figure them out quickly.

In some files the end of line/new line at the end is missing. I heard that this sometimes creates issues, so I guess it would be best to add them always.

cpp/models/abm/model_functions.cpp
```diff
@@ -0,0 +1,149 @@
...
+    if (person.get_location() != destination.get_id()) {
+        person.set_location(destination.get_id());
+        person.get_cells() = cells;
+        person.set_last_transport_mode(mode);
+
+        return true;
+    }
+    else {
+        return false;
+    }
```
khoanguyen-dev, line 145: Please add one test case here to cover the false scenario.

cpp/models/abm/world.cpp
```diff
@@ -38,20 +34,37 @@ namespace abm
...
+    assert(person.get_location().index != INVALID_LOCATION_INDEX);
+    assert(person.get_location().index < m_locations.size());
+    assert(person.get_age().get() < parameters.get_num_groups());
+
+    PersonId new_id{static_cast<uint32_t>(m_persons.size())};
+    m_persons.emplace_back(person, new_id);
+    auto& new_person = m_persons.back();
+    new_person.set_assigned_location(m_cemetery_id);
+
+    if (m_is_local_population_cache_valid) {
```
khoanguyen-dev, line 64: Please add a test to cover if `m_is_local_population_cache_valid`

cpp/models/abm/world.h
```diff
@@ -303,6 +303,133 @@ class World
...
+     * @brief Get the total number of Person%s at the Location.
+     * @param[in] location A LocationId from the world.
+     * @return Number of Person%s in the location.
+     */
+    size_t get_number_persons(LocationId location) const
+    {
+        if (!m_is_local_population_cache_valid) {
+            build_compute_local_population_cache();
+        }
+        return m_local_population_cache[location.index];
```
khoanguyen-dev, line 349: Please also add a test to cover `m_is_local_population_cache_valid`

cpp/models/abm/world.cpp
```diff
@@ -38,20 +34,37 @@ namespace abm
...
+        ++m_local_population_cache[new_person.get_location().index];
+    }
```
khoanguyen-dev, line 66: Please add a test for `m_is_local_population_cache_valid`

khoanguyen-dev (changes requested):
Thank you for making all the necessary changes. On my side, there are just some missing tests and comments.

cpp/models/abm/world.cpp
```diff
@@ -38,20 +34,37 @@ namespace abm
 LocationId World::add_location(LocationType type, uint32_t num_cells)
 {
     LocationId id = {static_cast<uint32_t>(m_locations.size()), type};
-    m_locations.emplace_back(std::make_unique<Location>(id, parameters.get_num_groups(), num_cells));
+    m_locations.emplace_back(id, parameters.get_num_groups(), num_cells);
```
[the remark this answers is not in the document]
reneSchm: Yes, emplace creates a new object from the given parameters (note that there is no initializer list used here), using a viable constructor. The old emplace used the move-ctor.

cpp/models/abm/world.cpp
```diff
@@ -65,32 +78,32 @@ void World::evolve(TimePoint t, TimeSpan dt)
...
 void World::interaction(TimePoint t, TimeSpan dt)
 {
+    const uint32_t num_persons = static_cast<uint32_t>(m_persons.size());
     PRAGMA_OMP(parallel for)
-    for (auto i = size_t(0); i < m_persons.size(); ++i) {
-        auto&& person     = m_persons[i];
-        auto personal_rng = Person::RandomNumberGenerator(m_rng, *person);
-        person->interact(personal_rng, t, dt, parameters);
+    for (uint32_t i = 0; i < num_persons; ++i) {
+        interact(i, t, dt);
```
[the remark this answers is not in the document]
reneSchm: In this case I would usually keep the `i`, since it is used in a very limited scope and doesn't do anything special here. But the loop in migration() should probably have a clearer iterator as it is used in a deeply nested context. I'll change both.

cpp/tests/test_abm_world.cpp
reneSchm: I think it would be a lot better to split the test. The idea for a unit test is to have a test case (i.e. both the "question" as well as the desired result) for a small unit of code. The evolve function mostly just calls other functions, so there is little to test there, instead, we should have a test `evolveBeginStepAspect1`, `evolveInteractAspect2`, and so on. Or skip the evolve part when we can access the "aspect" directly.

What is really important IMO, is that the test case has to be clear. The only thing I could tell you about "evolveMigration" is *that* it covers evolve. It is way to long for me to understand *how* without investing a lot of time. Also, but that is sadly true for most of our tests, it is unclear whether the test has an expected answer, or just a copy of the answer the code gave at the time of implementation (This is what I call a "consistency test". They can be useful, but you usually do not want one).

pycode/memilio-simulation/memilio/simulation/abm.cpp
```diff
@@ -196,11 +205,25 @@ PYBIND11_MODULE(_simulation_abm, m)
     pymio::bind_class<mio::abm::World, pymio::EnablePickling::Never>(m, "World")
         .def(py::init<int32_t>())
         .def("add_location", &mio::abm::World::add_location, py::arg("location_type"), py::arg("num_cells") = 1)
-        .def("add_person", &mio::abm::World::add_person, py::arg("location_id"), py::arg("age_group"),
-             py::return_value_policy::reference_internal)
-        .def_property_readonly("locations", &mio::abm::World::get_locations,
-                               py::keep_alive<1, 0>{}) //keep this world alive while contents are referenced in ranges
-        .def_property_readonly("persons", &mio::abm::World::get_persons, py::keep_alive<1, 0>{})
+        .def("add_person",
+             static_cast<mio::abm::PersonId (mio::abm::World::*)(mio::abm::LocationId, mio::AgeGroup)>(
```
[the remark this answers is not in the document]
reneSchm: Overloads. There are now two "add_person" methods, so we have to help the compiler decide which one to use. Using the function signature is an easy way to do so.

cpp/models/abm/world.cpp
```diff
@@ -38,20 +34,37 @@ namespace abm
...
-    assert(age.get() < parameters.get_num_groups());
-    uint32_t person_id = static_cast<uint32_t>(m_persons.size());
-    m_persons.push_back(std::make_unique<Person>(m_rng, get_individualized_location(id), age, person_id));
-    auto& person = *m_persons.back();
-    person.set_assigned_location(m_cemetery_id);
-    get_individualized_location(id).add_person(person);
-    return person;
+    assert(person.get_location().index != INVALID_LOCATION_INDEX);
+    assert(person.get_location().index < m_locations.size());
+    assert(person.get_age().get() < parameters.get_num_groups());
```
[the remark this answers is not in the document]
reneSchm: I think messages like that can be useful, but mostly on more complicated asserts, or when the message content could help avoid using a debugger. Since these are pretty simple (bound) checks, I don't think the messages are needed.

If you are undecided, then I would not add them. But I will, if you think they are beneficial.

cpp/models/abm/person.cpp
```diff
@@ -52,32 +51,17 @@ Person::Person(mio::RandomNumberGenerator& rng, Location& location, AgeGroup age
...
-        m_location->remove_person(*this);
-        m_location = &loc_new;
-        m_cells    = cells;
-        loc_new.add_person(*this, cells);
-        m_time_at_location    = TimeSpan(0);
-        m_last_transport_mode = transport_mode;
-    }
+    Person copy = *this;
+    copy.set_location(location);
+    return copy;
```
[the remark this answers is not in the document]
reneSchm: The function is definitely less dangerous now, and won't cause any side effects on the simulation.
And no, it is only used in a test and could be removed.

cpp/models/abm/location_type.h
```diff
@@ -57,7 +58,7 @@ static constexpr uint32_t INVALID_LOCATION_INDEX = std::numeric_limits<uint32_t>
  */
 struct LocationId {
     uint32_t index;
-    LocationType type;
+    LocationType type; // TODO: move to location
```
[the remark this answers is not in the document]
reneSchm: That is still an open question. On one hand, it makes sense to only use the index for the ID, as the type is not used for identification. On the other, whenever we need the index, we almost always want the type as well.

The first option is more semantically sound, the second is a lot more convenient (at least right now). I am not sure which is better.

cpp/models/abm/person.cpp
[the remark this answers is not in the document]
reneSchm: Should we remove copy_person? If so, do we remove Location::copy as well?

cpp/tests/test_abm_world.cpp
reneSchm: Exactly. I think we can trust the copy ctors. But if you do want to keep that test, we should wait for the world serialization s.t. we do not have to pseudo-implement "==" in the test.

cpp/models/abm/person.h
```diff
@@ -422,7 +422,8 @@ class Person
 
 private:
     LocationId m_location; ///< Current Location of the Person.
-    std::vector<uint32_t> m_assigned_locations; /**! Vector with the indices of the assigned Locations so that the 
+    LocationType m_location_type; ///< Type of the current Location.
```
reneSchm, line 425: This is not a change - it only makes explicit that the persons holds its locations type.

cpp/tests/test_abm_world.cpp
reneSchm: I opened #1067 so we can address this later. I'll remove the TODO, as it is out of scope for this issue.

cpp/models/abm/world.cpp
reneSchm: There is no meaningful way to test this line, as it is entirely internal model state. It rarely makes sense to test internal states, and this line's effect is basically impossible to observe, as is just a tiny optimization for when persons (but no locations!) get added after a begin_step has taken place.

I would honestly rather remove it and just invalidate the cache instead of writing a test just for this, since for now this optimization is completely unused. Or we accept that this is left uncovered.

cpp/models/abm/world.h
reneSchm: This check is there to cover the case where World::interact is called manually without a prior World::begin_step. The easy solution is to make interact private and remove this check.

And since this is purely internal state, we cannot test it, only cover it. And IMO that is a bad idea.

cpp/tests/test_json_serializer.cpp
```diff
@@ -487,85 +487,82 @@ TEST(TestJsonSerializer, abmLocation)
...
-    expected_json["age"]               = Json::UInt(2);
-    expected_json["id"]                = Json::UInt(person.get_person_id());
-    ASSERT_EQ(js.value(), expected_json);
-
-    // auto r = mio::deserialize_json(expected_json, mio::Tag<mio::abm::Person>());
-    // ASSERT_THAT(print_wrap(r), IsSuccess());
-    // EXPECT_EQ(r.value(), person);
-}
+// TEST(TestJsonSerializer, abmPerson) // FIXME: (de)serialize is only partially implemented
+// {
```
reneSchm, line 491: I commented these tests out for now, as ABM serialization was broken before this issue started, and a fix is already present in #652.

DavidKerkmann (commented):
Thanks for the changes! One idea:
We could omit the parameter LocationType in a lot of places where it is only necessary to identify the location, as the Id is unique for a location. Of course, this needs to stay for functions like get/set_assigned_location.

cpp/models/abm/person.h
```diff
@@ -163,26 +163,30 @@ class Person
     }
 
     /**
-     * @brief Set an assigned Location of the Person. 
+     * @brief Set an assigned Location of the Person.
+     *
+     * Important: Setting incorrect values will cause issues during simulation. It is preferable to use
+     *            World::assign_location with a valid LocationId, obtained e.g. through World::add_location.
```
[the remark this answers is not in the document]
reneSchm: This comment is supposed to replace the TODO from the .cpp file.
One issue would be to set an id to a value larger than m_locations.size(). Setting "wrong" types could also cause other issues, I assume during migration. So there will definitely be issues during simulation, but the person cannot verify anything.
To your second question - maybe? The behavior did not change here, we just do not package id/index and type together anymore. But we sometimes assume that the assigned id is valid and `type == m_locations[id].type`, hence the warning and mention of world.assign_location.

cpp/models/abm/world.cpp
khoanguyen-dev: Yes, I can see it's not covered also in the main branch. I guess we can leave it for now.

cpp/models/abm/person.h
reneSchm: Currently it is necessary. Otherwise, the functions using it (mainly migration rules) require the type or World&. But I claim this is out of scope and should be done in another issue (if you want that).

cpp/models/abm/testing_strategy.cpp
```diff
@@ -175,15 +178,15 @@ bool TestingStrategy::run_strategy(PersonalRandomNumberGenerator& rng, Person& p
...
     //lookup in std::vector instead of std::map should be much faster unless for large numbers of schemes
-    for (auto loc_key : {LocationId{location.get_index(), location.get_type()},
-                         LocationId{INVALID_LOCATION_INDEX, location.get_type()}}) {
+    for (auto key : {std::make_pair(location.get_type(), location.get_id()),
+                     std::make_pair(location.get_type(), LocationId::invalid_id())}) {
         auto iter_schemes =
-            std::find_if(m_location_to_schemes_map.begin(), m_location_to_schemes_map.end(), [loc_key](auto& p) {
-                return p.first == loc_key;
+            std::find_if(m_location_to_schemes_map.begin(), m_location_to_schemes_map.end(), [&](const auto& p) {
+                return p.type == key.first && p.id == key.second;
```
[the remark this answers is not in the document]
reneSchm: What do you propose?
Since it is very local, the short name is fine by me.

cpp/models/abm/location.h
```diff
@@ -96,48 +95,26 @@ struct Cell {
...
-        : Location(LocationId{loc_index, loc_type}, num_agegroups, num_cells)
-    {
-    }
+    explicit Location(LocationType loc_type, LocationId loc_id, size_t num_agegroups = 1, uint32_t num_cells = 1);
 
     /**
-     * @brief Return a copy of the current Location object.
-     * @param[in] other The original #Location.
+     * @brief Construct a copy of a Location with a new ID.
+     * @param[in] other A Location.
```
khoanguyen-dev, line 109:
```suggestion
     * @param[in] other The Location to be copied.
```

cpp/models/abm/location.h
```diff
@@ -96,48 +95,26 @@ struct Cell {
...
-     * @param[in] num_agegroups [Default: 1] The number of age groups in the model.
-     * @param[in] num_cells [Default: 1] The number of Cell%s in which the Location is divided.
-     */
-    Location(LocationId loc_id, size_t num_agegroups = 1, uint32_t num_cells = 1);
-
     /**
      * @brief Construct a Location with provided parameters. 
      * @param[in] loc_type The #LocationType.
-     * @param[in] index The index of the Location.
+     * @param[in] loc_id The index of the Location in the World.
```
khoanguyen-dev, line 101:
```suggestion
     * @param[in] loc_id The LocationId of the Location in the World.
```

cpp/models/abm/location.h
```diff
@@ -157,73 +134,22 @@ class Location
...
     LocationType get_type() const
     {
-        return m_id.type;
+        return m_type;
     }
 
     /**
-     * @brief Get the index of this Location.
-     * @return The index of the Location.
+     * @brief Get the location's identifier in a World.
```
khoanguyen-dev, line 141:
```suggestion
     * @brief Get the location's identifier in the World.
```

cpp/models/abm/location.h
```diff
@@ -396,12 +281,9 @@ class Location
     }
 
 private:
-    std::mutex m_mut; ///< Mutex to protect the list of persons from concurrent modification.
-    LocationId m_id; ///< Id of the Location including type and index.
-    bool m_capacity_adapted_transmission_risk; /**< If true considers the LocationCapacity for the computation of the 
-    transmission risk.*/
+    LocationType m_type; ///< Type of the Location.
+    LocationId m_id; ///< Id of the Location. Set by the World owning it.
```
khoanguyen-dev, line 285:
```suggestion
    LocationId m_id; ///< LocationId of the Location. Set by the World owning it.
```

cpp/models/abm/location.h
khoanguyen-dev: I also suggested to start using the name 'Model' instead of World.

cpp/models/abm/trip_list.h
```diff
@@ -38,7 +34,7 @@ namespace abm
  * @brief A trip describes a migration from one Location to another Location.
  */
 struct Trip {
-    uint32_t person_id; /**< Person that makes the trip and corresponds to the index into the structure m_persons from
+    PersonId person_id; /**< Person that makes the trip and corresponds to the index into the structure m_persons from
```
khoanguyen-dev, line 37:
```suggestion
    PersonId person_id; /**< PersonID of the Person that makes the trip and corresponds to the index into the structure m_persons from
```

cpp/models/abm/person.h
```diff
@@ -248,35 +161,32 @@ class Person
...
+     *
      * The assigned Location is saved by the index of its LocationId. Assume that a Person has at most one assigned
      * Location of a certain #LocationType.
      * @param[in] id The LocationId of the Location.
```
khoanguyen-dev, line 173:
```suggestion
     * @param[in] type The LocationType of the Location.
     * @param[in] id The LocationId of the Location.
```

cpp/models/abm/world.h
```diff
@@ -148,75 +151,78 @@ class World
...
      * @param[in] type Type of Location to add.
      * @param[in] num_cells [Default: 1] Number of Cell%s that the Location is divided into.
      * @return Index and type of the newly created Location.
      */
     LocationId add_location(LocationType type, uint32_t num_cells = 1);
 
-    /** 
+    /**
      * @brief Add a Person to the World.
      * @param[in] id Index and type of the initial Location of the Person.
```
khoanguyen-dev, line 178:
```suggestion
     * @param[in] id LocationId of the initial Location of the Person.
```

cpp/models/abm/world.h
```diff
@@ -148,75 +151,78 @@ class World
...
      * @param[in] age AgeGroup of the person.
      * @return Reference to the newly created Person.
      */
-    Person& add_person(const LocationId id, AgeGroup age);
+    PersonId add_person(const LocationId id, AgeGroup age);
+
+    // adds a copy of person to the world
```
khoanguyen-dev, line 184:
```suggestion
    /**
     * @brief Adds a copy of a given Person to the World.
     * @param[in] person The Person to be copied.
     * @return PersonId of the new Person.
     */
```

khoanguyen-dev (changes requested):
There are some comments that need to be updated. Otherwise, except for David's comments, it all looks good to me.

cpp/models/abm/location.h
reneSchm: "The index ..." does give you additional information on what loc_id is, while "The LocationId ..." just names the type again, are you sure this should be replaced?

And while the rename World to Model is planned, it is out of scope for this issue, and I will not partially rename it now.

cpp/models/abm/trip_list.h
reneSchm: I don't think repeating the type is helpful in these comments. They should explain/define what the variable does, using its type for that seems tautological. Also, its type is (usually) explicitly written right next to its name and definition anyways.

Although, if you think the comment lacks some important detail, we can update it.

cpp/models/abm/location.h
reneSchm: I changed it to "Unique identifier for the Location in the World owning it" instead?

reneSchm:
> We could omit the parameter LocationType in a lot of places where it is only necessary to identify the location, as the Id is unique for a location. 

My IDE usually complaints if there are unused variables, so I don't think that happens too often in our code. But we probably could rework e.g. the testing strategy, so that it stores two vectors, one for location id and one for type. This way we could also remove all the find-ifs and use constant time lookup instead.
But again, I would prefer opening a new issue instead of doing this in here as well.

cpp/models/abm/location.h
khoanguyen-dev: That's true. I guess "index" or "Unique identifier for the Location in the World" could also work. However, please be consistent with the naming in all comments.

cpp/models/abm/location.h
khoanguyen-dev: Same as above.

cpp/models/abm/location.h
reneSchm: What do you mean by consistent naming? Do you want all LocationId params to have the exact same doc comment?

Not sure that is necessary, but what about "Unique identifier for the Location in a World. Mostly obtained from World::add_location.".

cpp/models/abm/testing_strategy.cpp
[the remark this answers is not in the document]
DavidKerkmann: Short name is fine, it's just that `p` is usually used for persons. Perhaps `loc`?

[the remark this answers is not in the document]
reneSchm: `loc` is not really accurate as well (and neither is/was `m_location_to_schemes_map`), since this is a triple of loc id, loc type and a vector of schemes. How about something generic like `x`? Otherwise I would consider reworking the internals here in a new issue (also see my comment below on omitting loc types).

cpp/models/abm/location.h
khoanguyen-dev: > What do you mean by consistent naming? Do you want all LocationId params to have the exact same doc comment?
Not exactly, just sometimes loc_id is preferred as "Id of the Location", sometimes as "Index of the Location", or "location's identifier".  Maybe choose one of "identifier" or "index"?

cpp/models/abm/trip_list.h
khoanguyen-dev: Ok, I guess the part "corresponds to the index into the structure m_persons from" could be also a hint to the variable type already.

cpp/models/abm/location.h
reneSchm: To be honest, all of those are correct, and we state e.g. "The assigned Location is saved by the index of its LocationId", so it is not a secret either. ID is also just a shorthand for identifier.

More importantly, my changes here do not cover several instances of uses of id/index, so if you really want a unified description, you should open a new issue for this.

DavidKerkmann (approved):

cpp/models/abm/testing_strategy.cpp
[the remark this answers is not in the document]
DavidKerkmann: Right, perhaps `item` or `entry`? I'm fine with a separate issue.

cpp/models/abm/location.h
khoanguyen-dev: Ok, I guess we can leave them now and mention it in a future issue when all the documentation is updated.

khoanguyen-dev (approved):

DavidKerkmann (approved):

xsaschako:
Quick question:

In the benchmark get_subpopulation_combined is very slow, this maybe needs to be investigated.

reneSchm:
> In the benchmark get_subpopulation_combined is very slow, this maybe needs to be investigated.

This is partially intentional, as get_subpopulation needs to iterate over all persons instead of all persons at the given location, which we currently do not cache. Although get_subpopulation_combined is in fact a lot slower than it has to be right now, instead of calling get_subpopulation it should iterate over all persons exactly once. Something like:

```cpp
size_t World::get_subpopulation_combined(TimePoint t, InfectionState s) const
{
    return std::count_if(m_persons.begin(), m_persons.end(), [&](const auto& p) {
        return p.get_infection_state(t) == s;
    });
}
```

xsaschako:
Thank you! I mean i don't think its high priority, but as it is in the benchmark debuggging, it becomes unusable. Maybe we just think about a better way to debug the benchmark or implement this. but again, very low priority just something i noticed

PR #870, opened 2023-12-15
Run unit tests with OpenMP in CI

### LINKED ISSUE #786: Multithreaded Unittests for ABM

The ABM is parallelized using OpenMP but the Unit Tests are only executed single threaded. So there is no test that parallelized ABM produces the same results as sequential.

### Enhancement description

* Add a multithreaded runner
* Add a CI job that runs the unit tests with OPM_NUM_THREADS > 1, exact number depends on unit tests and runner

### Additional context

_No response_

### Checklist

- [X] Linked to project

dabele:
# Changes and Information

Add a job to the main CI pipeline that runs the unit tests with OpenMP enabled. 

simplified build artifact names in the CI pipeline, i.e. leave out default values.

The job runs with 4 threads. Github runners have 2 CPUs, I wasn't able to find information on Hyperthreading, I used 4 just in case. This is not very high parallelization, we may want to add a custom runner (maybe the HPDA cluster or something) to increase the chance of finding bugs. 

The unit tests are repeated 10 times, this was enough to reliably catch the segfault in #857, which happens ~50% of the time. That bug is quite easy to reproduce, we may want to add more repetitions in the future to catch rarer bugs.

So far I didn't add new unit tests just for OpenMP, the current tests catch race conditions in the currently parallelized parts. I don't know if there is a way to ensure 100% coverage of possible current and future race conditions. I think good general test coverage and maybe running testing loops a few more times than strictly necessary (e.g. do more simulation steps) is as good as it gets.

Multithreading issues are usually random, so it's unclear what the failure will be. On my local machine, one test went into an infinite loop sometimes (due to the bug in #857), so I added a timeout to the test. Otherwise the test would keep spinning for 6 hours (the github limit). The tests might also just work as expected even if there is a race condition. To increase the chance of catching a race condition, I added a reminder to the pull request template to run unit tests with OpenMP locally.

Closes #786 

### Checks by code author

- [x] (For ABM development) Checked [benchmark results](https://github.com/SciCompMod/memilio/wiki/Agent-Based-Model-Development)

### DIFF
--- a/.github/actions/linux-build/action.yml
+++ b/.github/actions/linux-build/action.yml
@@ -37,6 +37,10 @@ inputs:
     description: "Build benchmarks"
     required: false
     default: "OFF"
+  enable-openmp:
+    description: "Enable Multithreading with OpenMP"
+    required: false
+    default: "OFF"
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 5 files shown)


dabele:
merged main branch, openmp unittest is passing now as expected.

dabele:
> good changes, looks good actually i think github upgraded the runners to be 4 cored for public repositories so you may want to change it to 8 , see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners

Since we already have 4 threads enabled and don't know about hyperthreading this should be fine for now, 4 are probably enough to catch issues.

PR #960, opened 2024-02-28
949 Correct handling of dt_min in integrators

### LINKED ISSUE #949: Correct handling of dt_min in ControlledStepperWrapper
### Bug description

- ControlledStepperWrapper incorrectly returns a failed step size adaption if `dt==dt_min`
- ControlledStepperWrapper may set dt to a value smaller than dt_min

### Version

Any

### To reproduce

use a ControlledStepperWrapper for example with `dt_max==dt_min`

### Relevant log output

_No response_

_No response_

### Checklist

- [x] Linked to project

reneSchm:
now the integrator makes a time step of at least dt_min, instead of potentially doing nothing except decreasing dt without limit. this breaks fsal steppers (in part. rk-dopri5, only used in tests and benchmarks), so they were disabled for now.

# Changes and Information

- made changes to ControlledStepperWrapper such that the step method
  - correctly uses dt_min
  - returns a dt in [dt_min, dt_max]
  - always makes an integration step of at least size dt_min, even if the step got rejected due to not reaching tolerances
- updated RKIntegratorCore to behave similarly. 

- Please check that the requirements listed in the documentation of IntegratorCore::step are understandable and sufficient.
- Also consider whether the compare-with tests are correctly updated, or if the testing method might need to change.

### Checks by code author

Closes #949


reneSchm:
I fixed the issue using another version of the try_step method. However, this breaks the runge_kutta_dopri5 stepper due to a [bug in boost](https://github.com/boostorg/odeint/issues/72). Since we currently only use it in testing and benchmarks, I decided to prevent it (and other FSAL steppers) to be used in the wrapper.

@mknaranja Do you think its okay to disable FSAL steppers for now? Otherwise, the ControlledStepperWrapper would have to make a manual integration step whenever the normal try_step loop fails. This is not a terrible solution, but we potentially make a repeated integration step whenever dt==dt_min.

The test TestOdeSecir.compareAgeResWithPreviousRunCashKarp currently fails since it uses an integration result with a time step < dt_min as reference. @MaxBetzDLR is it safe to just update the csv with a new integration result?

reneSchm:
Currently, both ControlledStepperWrapper and RKIntegratorCore are allowed to make steps of size > dt_max, only if the function argument dt > dt_max.
We could either keep this behaviour, I don't think it will matter much or have a negative impact. But it may be unexpected, so we could warn/exit on a bad input; or allow any input value, but restrict it to [dt_min, dt_max] internally.
I would prefer restricting dt internally, as we already enforce the lower bound of dt_min.

mknaranja:
> Currently, both ControlledStepperWrapper and RKIntegratorCore are allowed to make steps of size > dt_max, only if the function argument dt > dt_max. We could either keep this behaviour, I don't think it will matter much or have a negative impact. But it may be unexpected, so we could warn/exit on a bad input; or allow any input value, but restrict it to [dt_min, dt_max] internally. I would prefer restricting dt internally, as we already enforce the lower bound of dt_min.

@reneSchm This is not good design we did. We should not allow that, adapt it and inform/warn on it.

cpp/memilio/math/stepper_wrapper.h
```diff
@@ -70,32 +76,59 @@ class ControlledStepperWrapper : public mio::IntegratorCore
...
+        // make a integration step, adapting dt to a possibly larger value on success,
+        // or a strictly smaller value on fail.
+        // stop only on a successful step or a failed step size adaption (w.r.t. the minimal step size m_dt_min)
+        while (step_result == fail && is_dt_valid) {
+            if (dt < m_dt_min) {
+                is_dt_valid = false;
+                dt          = m_dt_min;
+            }
+            // we use the scheme try_step(sys, in, t, out, dt) with sys=f, in=y(t_{k}), out=y(t_{k+1}).
+            // this is similiar to do_step, but it can adapt the step size dt. if successfull, it also updates t.
```
mknaranja, line 106:
```suggestion
            // this is similiar to do_step, but it can adapt the step size dt. If successful, it also updates t.

```

cpp/memilio/math/stepper_wrapper.h
```diff
@@ -70,32 +76,59 @@ class ControlledStepperWrapper : public mio::IntegratorCore
...
+                        dxds.resizeLike(x); // boost resizers cannot resize Eigen::Vector, hence we need to do that here
+                        f(x, s, dxds);
+                    },
+                    m_yt, t, m_ytp1, dt);
+            }
+        }
+        // output the new value by copying it back to the reference
+        ytp1 = m_ytp1;
+        // bound dt from below
+        // the last adaptive step (successfull or not) may have calculated a new step size smaller than m_dt_min
```
mknaranja, line 120:
```suggestion
        // the last adaptive step (successful or not) may have calculated a new step size smaller than m_dt_min

```

mknaranja (commented):
Hi @reneSchm 
Thank you very much. I had looked at almost everything. I have mostly commenting and documentation suggestions. If someone could have a more detailed look on the new `step` method, that would be great though. If that's done, we can merge it.

cpp/memilio/math/adapt_rk.cpp
```diff
@@ -73,6 +73,10 @@ Tableau::Tableau()
 bool RKIntegratorCore::step(const DerivFunction& f, Eigen::Ref<const Eigen::VectorXd> yt, double& t, double& dt,
                             Eigen::Ref<Eigen::VectorXd> ytp1) const
 {
+    assert(0 <= m_dt_min);
+    assert(m_dt_min < m_dt_max);
```
jubicker, line 77: should m_dt_min = m_dt_max also be possible if we want to force a fixed step size?

cpp/memilio/math/integrator.h
```diff
@@ -43,13 +40,25 @@ class IntegratorCore
...
+     * These changes are noted in (brackets) below.
+     * Adaptive integrators must have bounds dt_min and dt_max for dt.
+     * The adaptive step sizing is considered to have failed, if dt would be adapted to a value strictly below dt_min.
+     * Even if the step sizing failed, the integrator must make a step of at least size dt_min.
+     *
+     * @param[in] f Right hand side of ODE. May be called multiple times with different arguments.
+     * @param[in] yt The known value of y at time t.
+     * @param[in,out] t The current time. It will be increased by dt.
+     *     (If adaptive, the increment is from [dt_min, dt].)
+     * @param[in,out] dt The current step size h=dt.
```
jubicker, line 55: Is it the current or the next step size? Below it is said that dt is an estimate of the next step, so it should be the next step size, rights?

cpp/memilio/math/stepper_wrapper.h
```diff
@@ -70,32 +76,59 @@ class ControlledStepperWrapper : public mio::IntegratorCore
...
-                    f(x, s, dxds);
-                },
-                dydt, t, dt);
-            // stop on a successfull step or a failed step size adaption (w.r.t. the minimal step size)
-        } while (t == t_old && dt > m_dt_min);
-        ytp1 = dydt; // output new y(t)
-        return dt > m_dt_min;
+        using boost::numeric::odeint::fail;
+        assert(0 <= m_dt_min);
+        assert(m_dt_min <= m_dt_max);
```
jubicker, line 90: Above you assert  m_dt_min < m_dt_max

cpp/tests/test_numericalIntegration.cpp
```diff
@@ -139,6 +139,69 @@ TYPED_TEST(TestVerifyNumericalIntegrator, sine)
     EXPECT_NEAR(this->err, 0.0, 1e-7);
 }
 
+TYPED_TEST(TestVerifyNumericalIntegrator, adaptiveStepSizing)
+{
+    // set large tolarances and big dt_min, since we want that the step size adaption fails with dt_min
```
jubicker, line 144:
```suggestion
    // set fixed tolerances and big dt_min, since we want that the step size adaption fails with dt_min
```

cpp/benchmarks/simulation.cpp
```diff
@@ -49,8 +49,8 @@ BENCHMARK_TEMPLATE(simulation, mio::RKIntegratorCore)->Name("Dummy 3/3");
 BENCHMARK_TEMPLATE(simulation, mio::RKIntegratorCore)->Name("simulate SecirModel adapt_rk");
 BENCHMARK_TEMPLATE(simulation, mio::ControlledStepperWrapper<boost::numeric::odeint::runge_kutta_cash_karp54>)
     ->Name("simulate SecirModel boost rk_ck54");
-BENCHMARK_TEMPLATE(simulation, mio::ControlledStepperWrapper<boost::numeric::odeint::runge_kutta_dopri5>)
-    ->Name("simulate SecirModel boost rk_dopri5");
+// BENCHMARK_TEMPLATE(simulation, mio::ControlledStepperWrapper<boost::numeric::odeint::runge_kutta_dopri5>)
+//     ->Name("simulate SecirModel boost rk_dopri5"); // TODO: reenable once boost bug is fixed
```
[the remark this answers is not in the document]
reneSchm: There is now #966

cpp/memilio/math/integrator.h
```diff
@@ -43,13 +40,25 @@ class IntegratorCore
...
-     * @brief Step of the integration with possible adaptive with
+     * @brief Make a single integration step.
      *
-     * @param[in] f Right hand side of ODE
-     * @param[in] yt value of y at t, y(t)
-     * @param[in,out] t current time step h=dt
-     * @param[in,out] dt current time step h=dt
-     * @param[out] ytp1 approximated value y(t+1)
+     * The behaviour of this method changes slightly when the integration scheme has adaptive step sizing. 
+     * These changes are noted in (brackets) below.
```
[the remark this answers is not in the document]
reneSchm: Aren't those braces?

cpp/memilio/math/integrator.h
```diff
@@ -43,13 +40,25 @@ class IntegratorCore
...
+     * Adaptive integrators must have bounds dt_min and dt_max for dt.
+     * The adaptive step sizing is considered to have failed, if dt would be adapted to a value strictly below dt_min.
+     * Even if the step sizing failed, the integrator must make a step of at least size dt_min.
+     *
+     * @param[in] f Right hand side of ODE. May be called multiple times with different arguments.
+     * @param[in] yt The known value of y at time t.
+     * @param[in,out] t The current time. It will be increased by dt.
+     *     (If adaptive, the increment is from [dt_min, dt].)
```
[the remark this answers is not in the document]
reneSchm: No, the integrator never makes a bigger time step than dt. To do that it would have to repeat a successful step again, which is inefficient (and I think may loose accuracy?).

cpp/memilio/math/adapt_rk.cpp
reneSchm: Yes, good catch. This should be <=

cpp/memilio/math/integrator.h
reneSchm: I'll need rephrase this - dt [in] is the current step size, dt [out] is _set to_ the next. The latter does not happen with non-adaptive steppers.

cpp/tests/test_numericalIntegration.cpp
```diff
@@ -139,6 +139,69 @@ TYPED_TEST(TestVerifyNumericalIntegrator, sine)
...
+    t_eval    = 0;
+    this->dt  = dt_min;
+    step_okay = integrator.step(
+        [](const auto&&, auto&&, auto&& dxds) {
+            dxds.setZero(); // const f
+        },
+        this->y[0], t_eval, this->dt, this->sol[0]);
+
+    EXPECT_EQ(step_okay, true); // step sizing must be okay
+    EXPECT_GE(this->dt, dt_min); // new step size may be larger
```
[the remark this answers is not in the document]
reneSchm: That case gets covered by the check below. I don't think double checking is necessary.
But I'll think about rewriting these checks to always test all bounds all the time.

HenrZu (commented):
I looked at the step functions and they look correct and also well commented:)
One comment. Should we also adjust the default values for the minimum step sizes? (also in stepper_wrapper.h) 

https://github.com/SciCompMod/memilio/blob/afd49822b2bcba997b32d770ca8eead2be0faaa3/cpp/memilio/math/adapt_rk.h#L108

reneSchm:
> One comment. Should we also adjust the default values for the minimum step sizes? (also in stepper_wrapper.h)

I don't think we should set a different dt_min, which scale would we even choose?

Maybe we could set a dt_min in the simulation, but I would prefer to keep it model/executable specific.

reneSchm:
So I added a small warning when the given dt was out of bounds, and strangely enough it might have improved performance for some steppers?
- `RKIntegratorCore` aka adapt_rk and  `ControlledStepperWrapper<runge_kutta_fehlberg78>` aka rkf78 are consistently about 50ns faster. Which is nice, but confusing. Apart from the couple lines for the logging, the assembly looked like only some addresses got reordered, at least I didn't see any significant changes. My best guess is that I got lucky and some reordering caused better cache hits, but I don't know.
- `ControlledStepperWrapper<runge_kutta_cash_carp54>` aka rk_ck54 seems to be roughly 10ns slower, but that is within the margin of error for these measurements. This is as expected, and I think well acceptable.

Before logging (dae153c):
```
simulate SecirModel adapt_rk            1008 ns         1007 ns       696182
simulate SecirModel boost rk_ck54        746 ns          746 ns       938309
simulate SecirModel boost rkf78         2196 ns         2195 ns       317635
```
With logging (b93c0e6):
```
simulate SecirModel adapt_rk             935 ns          934 ns       749251
simulate SecirModel boost rk_ck54        759 ns          759 ns       921252
simulate SecirModel boost rkf78         2141 ns         2141 ns       323717
```

mknaranja (approved):

PR #971, opened 2024-03-19
55 Write mobility data

HenrZu:
# Changes and Information

-
-
-

-
-

### Checks by code author

- [ ] New code adheres to [coding guidelines](https://github.com/SciCompMod/memilio/wiki/Coding-guidelines)

### DIFF
--- a/cpp/memilio/mobility/metapopulation_mobility_instant.cpp
+++ b/cpp/memilio/mobility/metapopulation_mobility_instant.cpp
@@ -21,5 +21,26 @@
 
 namespace mio
 {
+void MigrationEdge::condense_m_migrated(const double t, const std::vector<size_t>& indices_non_symptomatic,
+                                        const std::vector<size_t>& indices_symptomatic)
+{
+
+    const auto& last_value = m_migrated.get_last_value();
+
+    auto num_INS =
+        std::accumulate(indices_non_symptomatic.begin(), indices_non_symptomatic.end(), 0, [&](auto sum, auto i) {
+            return sum + last_value[i];
+        });
+
+    auto num_ISy = std::accumulate(indices_symptomatic.begin(), indices_symptomatic.end(), 0, [&](auto sum, auto i) {
+        return sum + last_value[i];
+    });
+
+    double total_commuters = m_migrated.get_last_value().sum();
+
+    // as time point t which contains now the carriers, infectious and total over age groups
+    m_num_migrated.add_time_point(t,
+                                  (mio::TimeSeries<double>::Vector(3) << num_INS, num_ISy, total_commuters).finished());
+}
 
 } // namespace mio
--- a/cpp/memilio/mobility/metapopulation_mobility_instant.h
+++ b/cpp/memilio/mobility/metapopulation_mobility_instant.h
@@ -241,6 +241,41 @@ private:
     DynamicNPIs m_dynamic_npis;
 };
 
+/**
+ * detect a get_indices_of_symptomatic_and_nonsymptomatic function for the Model type.
+ */
+template <class Sim>
+using get_indices_of_symptomatic_and_nonsymptomatic_expr_t =
+    decltype(get_indices_of_symptomatic_and_nonsymptomatic(std::declval<Sim&>()));
+
+/**
+ * @brief Get the indices of symptomatic and non-symptomatic infection states.
+ *
+ * This function generates two vectors of indices, one for non-symptomatic infection states and one for symptomatic infection states.
+ * Each vector contains the flat indices of the corresponding infection states for each age group and compartment in the model.
+ *
+ * @tparam Base The base class for the simulation, defaults to mio::Simulation<Model>.
+ * @param[in] sim The simulation object from which we obtain the model.
+ *
+ * @return A tuple containing two vectors of size_t. The first vector contains the indices of non-symptomatic infection states,
+ * and the second vector contains the indices of symptomatic infection states. The indices are ordered first by age group, then by compartment.
+ */
+template <class Sim,
+          std::enable_if_t<!is_expression_valid<get_indices_of_symptomatic_and_nonsymptomatic_expr_t, Sim>::value,
+                           void*> = nullptr>
+auto get_indices_of_symptomatic_and_nonsymptomatic(SimulationNode<Sim>& /*node*/)
+{
+    return std::make_tuple(std::vector<size_t>{}, std::vector<size_t>{});
+}
+
+template <class Sim,
+          std::enable_if_t<is_expression_valid<get_indices_of_symptomatic_and_nonsymptomatic_expr_t, Sim>::value,
+                           void*> = nullptr>
+auto get_indices_of_symptomatic_and_nonsymptomatic(SimulationNode<Sim>& node)
+{
+    return get_indices_of_symptomatic_and_nonsymptomatic(node.get_simulation());
+}
+
 /** 
  * represents the migration between two nodes.
  */
@@ -256,6 +291,7 @@ public:
         , m_migrated(params.get_coefficients().get_shape().rows())
         , m_return_times(0)
         , m_return_migrated(false)
+        , m_num_migrated(3)
     {
     }
 
@@ -268,6 +304,7 @@ public:
         , m_migrated(coeffs.rows())
         , m_return_times(0)
         , m_return_migrated(false)
+        , m_num_migrated(3)
     {
     }
 
@@ -277,6 +314,19 @@ public:
     const MigrationParameters& get_parameters() const
     {
         return m_parameters;
+    }
+
+    /**
+    * Retrieve the count of commuters in the infection states: InfectedNoSymptoms and InfectedSymptomsNaive, 
+    * along with the total number of commuter.
+    */
+    TimeSeries<ScalarType>& get_migrated()
+    {
+        return m_num_migrated;
+    }
+    const TimeSeries<ScalarType>& get_migrated() const
+    {
+        return m_num_migrated;
     }
 
     /**
@@ -299,6 +349,16 @@ private:
     bool m_return_migrated;
     double m_t_last_dynamic_npi_check               = -std::numeric_limits<double>::infinity();
     std::pair<double, SimulationTime> m_dynamic_npi = {-std::numeric_limits<double>::max(), SimulationTime(0)};
+    TimeSeries<double> m_num_migrated;
+
+    /**
+     * Computes a condensed version of m_migrated and puts it in m_num_migrated.
+     * m_num_migrated then only contains commuters with infection states InfectedNoSymptoms and InfectedSymptoms.
+     * Additionally, the total number of commuters is stored in the last entry of m_num_migrated.
+     * @param[in] t current time
+     */
+    void condense_m_migrated(const double t, const std::vector<size_t>& indices_non_symptomatic,
+                             const std::vector<size_t>& indices_symptomatic);
 };
 
 /**
@@ -389,8 +449,8 @@ auto get_migration_factors(const SimulationNode<Sim>& node, double t, const Eige
  * detect a get_migration_factors function for the Model type.
  */
 template <class Sim>
-using test_commuters_expr_t = decltype(test_commuters(
-    std::declval<Sim&>(), std::declval<Eigen::Ref<const Eigen::VectorXd>&>(), std::declval<double>()));
+using test_commuters_expr_t = decltype(
+    test_commuters(std::declval<Sim&>(), std::declval<Eigen::Ref<const Eigen::VectorXd>&>(), std::declval<double>()));
 
 /**
  * Test persons when migrating from their source node.
@@ -484,6 +544,10 @@ void MigrationEdge::apply_migration(double t, double dt, SimulationNode<Sim>& no
 
         node_to.get_result().get_last_value() += m_migrated.get_last_value();
         node_from.get_result().get_last_value() -= m_migrated.get_last_value();
+
+        static auto indices_tuple                     = get_indices_of_symptomatic_and_nonsymptomatic(node_from);
+        auto& [indices_no_symptoms, indices_symptoms] = indices_tuple;
+        condense_m_migrated(t, indices_no_symptoms, indices_symptoms);
     }
     m_return_migrated = !m_return_migrated;
 }
--- a/cpp/models/ode_secir/model.h
+++ b/cpp/models/ode_secir/model.h
@@ -29,6 +29,7 @@
 #include "memilio/math/smoother.h"
 #include "memilio/math/eigen_util.h"
 #include "memilio/math/interpolation.h"
+#include <cstddef>
 
 namespace mio
 {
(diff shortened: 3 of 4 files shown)


HenrZu:
Using the setup in our simulation with ~30.000 edges and 50 simulation days, the size of the Edges.h file is 135mb.

HenrZu:
There is an inconsistency in our project between the use of the terms migrated and mobility. I would leave them here for now and edit this PR after merging so that we only use mobility.

cpp/memilio/mobility/metapopulation_mobility_instant.h
```diff
@@ -302,8 +391,44 @@ class MigrationEdge
...
+    std::vector<std::vector<size_t>> m_save_indices; // groups of indices from compartments to save
+    TimeSeries<double> m_mobility_results; // save results from edges + entry for the total number of commuters
+
+    /**
+     * Computes a condensed version of m_migrated and puts it in m_mobility_results.
+     * m_mobility_results then only contains commuters with infection states InfectedNoSymptoms and InfectedSymptoms.
+     * Additionally, the total number of commuters is stored in the last entry of m_mobility_results.
+     * @param[in] t current time
+     */
+    void condense_m_mobility(const double t);
```
reneSchm, line 403: I would much prefer if this mobility result could be computed outside of this class in a post-processing. As far as I can see, that *is* exactly how m_mobility_results is currently being used. Or is it (supposed to be) used anywhere during a simulation? If not, I propose that we use a free function (e.g. "condense_mobility_result" or "accumulate_compartments") that takes a TimeSeries (e.g. m_migrated) as well as those indices and returns a condensed TimeSeries, instead of the class member. This solution is a lot more flexible, e.g. it allows multiple "condensations" in one simulation, and the condense function could be reused for non-mobility results.

cpp/memilio/io/result_io.cpp
```diff
@@ -89,6 +89,118 @@ IOResult<void> save_result(const std::vector<TimeSeries<double>>& results, const
...
+            H5DataSpace dspace_values{H5Screate_simple(2, dims_values, NULL)};
+            MEMILIO_H5_CHECK(dspace_values.id, StatusCode::UnknownError, "Values DataSpace could not be created.");
+            auto dset_name = "End" + std::to_string(ids[edge_indx].second);
+            H5DataSet dset_values{H5Dcreate(start_node_h5group.id, dset_name.c_str(), H5T_NATIVE_DOUBLE,
+                                            dspace_values.id, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT)};
+            MEMILIO_H5_CHECK(dset_values.id, StatusCode::UnknownError, "Values DataSet could not be created.");
+
+            MEMILIO_H5_CHECK(
+                H5Dwrite(dset_values.id, H5T_NATIVE_DOUBLE, H5S_ALL, H5S_ALL, H5P_DEFAULT, edge_result.data()),
+                StatusCode::UnknownError, "Values data could not be written.");
```
reneSchm, line 187: The error messages in this function are not very helpfull, if you don't already know what you are looking for. Can you expand these?

cpp/memilio/io/result_io.h
```diff
@@ -121,6 +121,29 @@ IOResult<void> save_result_with_params(const std::vector<TimeSeries<double>>& re
...
+ * @param result Simulation results per edge of the graph.
+ * @param ids Identifiers for the start and end node of the edges.
+ * @param filename Name of file
+ * @return Any io errors that occur during writing of the files. 
+ */
+IOResult<void> save_edges(const std::vector<TimeSeries<double>>& results, const std::vector<std::pair<int, int>>& ids,
+                          const std::string& filename);
+
+/**
+ * Save the results for the Edges obtained from the function condense_m_mobility.
```
reneSchm, line 135: You cannot use the name of a private member of an unnamed class in the documentation here. No one will be able to understand this connection

cpp/memilio/io/result_io.h
```diff
@@ -121,6 +121,29 @@ IOResult<void> save_result_with_params(const std::vector<TimeSeries<double>>& re
...
+ * @param ensemble_edges Simulation results for each run for each edge.
+ * @param pairs_edges Identifiers for the start and end node of the edges.
+ * @param result_dir Top level directory for all results of the parameter study.
+ * @param save_single_runs [Default: true] Defines if single run results are written.
+ * @param save_single_runs [Default: true] Defines if percentiles are written.
```
reneSchm, line 140:
```suggestion
 * @param save_percentiles [Default: true] Defines if percentiles are written.
```
Just an idea, change this from bool to a default empty vector, so that the user can specify which percentiles are printed?

cpp/memilio/io/result_io.cpp
```diff
@@ -89,6 +89,118 @@ IOResult<void> save_result(const std::vector<TimeSeries<double>>& results, const
...
+            for (Eigen::Index t_idx = 0; t_idx < result_edge.get_num_time_points(); ++t_idx) {
+                auto v                 = result_edge.get_value(t_idx).transpose().eval();
+                edge_result.row(t_idx) = v;
+                total.row(t_idx) += v;
+            }
+
+            hsize_t dims_values[] = {static_cast<hsize_t>(num_timepoints), static_cast<hsize_t>(num_infectionstates)};
+            H5DataSpace dspace_values{H5Screate_simple(2, dims_values, NULL)};
+            MEMILIO_H5_CHECK(dspace_values.id, StatusCode::UnknownError, "Values DataSpace could not be created.");
+            auto dset_name = "End" + std::to_string(ids[edge_indx].second);
```
reneSchm, line 180: What is "End"? Is it a Group total (over time)?

cpp/memilio/mobility/metapopulation_mobility_instant.cpp
```diff
@@ -18,8 +18,8 @@
 * limitations under the License.
 */
 #include "memilio/mobility/metapopulation_mobility_instant.h"
+#include "memilio/utils/compiler_diagnostics.h"
```
reneSchm, line 21: This include seems unnecessary, remove it?

cpp/memilio/mobility/metapopulation_mobility_instant.h
```diff
@@ -179,6 +206,20 @@ class MigrationParameters
...
+     */
+    const auto& get_save_indices() const
+    {
+        return m_save_indices;
+    }
+
+    const auto& get_save_indices()
+    {
+        return m_save_indices;
+    }
```
reneSchm, line 221: This function is superfluous, or did you mean to make the return type non-const?

cpp/tests/test_save_results.cpp
```diff
@@ -339,4 +370,129 @@ TEST(TestSaveResult, save_percentiles_and_sums)
...
+    auto pairs_edges = std::vector<std::pair<int, int>>{};
+    pairs_edges.push_back({0, 1});
+    pairs_edges.push_back({1, 0});
+
+    TempFileRegister file_register;
+    auto results_file_path = file_register.get_unique_path("test_result-%%%%-%%%%.h5");
+    auto save_edges_status = mio::save_edges(results_edges, pairs_edges, results_file_path);
+    ASSERT_TRUE(save_edges_status);
+
+    auto results_from_file = mio::read_result(results_file_path);
```
reneSchm, line 473: This test does not actually test whether save_edges works. At best, you assert that read_result can deal with what save_edges outputs. But, without making sure this output is correct, that does not mean much.

A better test would be to set up a few (mock) edges with small results as input, and then check whether those are saved as expected - i.e., compare them against a manually written string.
This setup can then also be used to test read_result, by exchanging input and output.

cpp/memilio/io/result_io.cpp
```diff
@@ -89,6 +89,118 @@ IOResult<void> save_result(const std::vector<TimeSeries<double>>& results, const
     }
     return success();
 }
+IOResult<void> save_edges(const std::vector<std::vector<TimeSeries<double>>>& ensemble_edges,
+                          const std::vector<std::pair<int, int>>& pairs_edges, const fs::path& result_dir,
+                          bool save_single_runs, bool save_percentiles)
```
reneSchm, line 94: Why do we need these functions? Since they also save a (collection of) TimeSeries, couldn't we use the save_result functions with a different file name?

reneSchm (changes requested):
I have not looked at everything yet, and we should first discuss whether the condensed edge result (see my other comment), but here are some more things I found.

cpp/memilio/io/result_io.h
```diff
@@ -121,6 +121,29 @@ IOResult<void> save_result_with_params(const std::vector<TimeSeries<double>>& re
...
+ * @param filename Name of file
+ * @return Any io errors that occur during writing of the files. 
+ */
+IOResult<void> save_edges(const std::vector<TimeSeries<double>>& results, const std::vector<std::pair<int, int>>& ids,
+                          const std::string& filename);
+
+/**
+ * Save the results for the Edges obtained from the function condense_m_mobility.
+ * @param ensemble_edges Simulation results for each run for each edge.
+ * @param pairs_edges Identifiers for the start and end node of the edges.
```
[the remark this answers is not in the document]
HenrZu: Could be both. Doesnt really matter. But in our classical use case it is used with the ids.

cpp/memilio/io/result_io.h
HenrZu: i think this is a good idea. But i will shift this to another issue

cpp/tests/test_save_results.cpp
HenrZu: adjusted the tests

cpp/memilio/mobility/metapopulation_mobility_instant.h
HenrZu: The main idea for this function was to test whether the indices were set correctly. Currently used in test_graph.

cpp/memilio/io/result_io.cpp
HenrZu: I tried that first. However, there were some problems. The ids are given by pairs and not by pure ints. In addition, the groups are now the start edges. The next level is then a group with the name "End", which denotes the end edge. These are not necessarily sequential as with age groups. (E.g. Could be 1002, 1009,.. and not 1,2,3,4...)
In addition, there were several small things that should have been adjusted.

 As the function is already very complex and difficult to understand, I decided to go without further abstractions and write a separate function for the edges.

cpp/memilio/mobility/metapopulation_mobility_instant.h
HenrZu: As already discussed, post-processing is not possible here, as we reset the results to the m_migrated after each day.
Therefore, the current solution is flexible within the scope of the possibilities. I have tried the idea of integrating a history object like in the advance function as already used in the ABM, but this requires a lot of changes, so I would leave it at the current version

cpp/memilio/mobility/metapopulation_mobility_instant.h
```diff
@@ -179,7 +206,14 @@ class MigrationParameters
     {
         m_coefficients = coeffs;
     }
-    /** @} */
+
+    /**
+     * @return the indices we want to save for the edge
+     */
```
reneSchm, line 212: Please update this alongside the constructor comment.

cpp/memilio/mobility/metapopulation_mobility_instant.h
```diff
@@ -129,6 +129,8 @@ class MigrationParameters
      */
     MigrationParameters(const MigrationCoefficientGroup& coeffs)
         : m_coefficients(coeffs)
+        , m_save_indices(0)
+
```
reneSchm, line 133: When you update the name above, please do it here as well. 

(And there is an additional newline)

cpp/memilio/mobility/metapopulation_mobility_instant.h
```diff
@@ -302,8 +384,44 @@ class MigrationEdge
...
+    std::vector<std::vector<size_t>> m_save_indices; // groups of indices from compartments to save
+    TimeSeries<double> m_mobility_results; // save results from edges + entry for the total number of commuters
+
+    /**
+     * Computes a condensed version of m_migrated and puts it in m_mobility_results.
+     * m_mobility_results then only contains commuters with infection states InfectedNoSymptoms and InfectedSymptoms.
+     * Additionally, the total number of commuters is stored in the last entry of m_mobility_results.
+     * @param[in] t current time
+     */
+    void condense_m_mobility(const double t);
```
reneSchm, line 396:
```suggestion
    void add_mobility_result_time_point(const double t);
```
Or something similar?

cpp/tests/test_save_results.cpp
```diff
@@ -339,4 +370,100 @@ TEST(TestSaveResult, save_percentiles_and_sums)
...
+    // group 0
+    auto result_from_file_group0 = results_from_file.value()[0];
+    ASSERT_EQ(result_from_file_group0.get_groups().get_num_time_points(), 3);
+    ASSERT_EQ(result_from_file_group0.get_groups().get_num_elements(), 6);
+    ASSERT_EQ(result_from_file_group0.get_groups().get_value(0), (Eigen::VectorXd(6) << 0, 0, 0, 2, 2, 2).finished());
+    ASSERT_EQ(result_from_file_group0.get_groups().get_value(1), (Eigen::VectorXd(6) << 1, 1, 1, 3, 3, 3).finished());
+    ASSERT_EQ(result_from_file_group0.get_groups().get_value(2), (Eigen::VectorXd(6) << 1, 1, 1, 5, 5, 5).finished());
+    ASSERT_EQ(result_from_file_group0.get_groups().get_time(0), 0.0);
+    ASSERT_EQ(result_from_file_group0.get_groups().get_time(1), 1.0);
+    ASSERT_EQ(result_from_file_group0.get_groups().get_time(2), 2.0);
```
reneSchm, line 444: Please only use ASSERT_X when the test needs to stop on failure. Multiple EXPECT_X means potentially a lot more output on failure, but you usually want that to find the error/bug that caused the test to fail.

reneSchm (changes requested):
There is still some documentation that is missing new arguments, briefs or the param flags (i.e. [in], [out], [inout]). Please also make sure to use full sentences.

cpp/memilio/io/result_io.h
```diff
@@ -121,6 +121,29 @@ IOResult<void> save_result_with_params(const std::vector<TimeSeries<double>>& re
...
+ * @param result_dir Top level directory for all results of the parameter study.
+ * @param save_single_runs [Default: true] Defines if single run results are written.
+ * @param save_percentiles [Default: true] Defines if percentiles are written.
+ * @return Any io errors that occur during writing of the files.
+ */
+IOResult<void> save_edges(const std::vector<std::vector<TimeSeries<double>>>& ensemble_edges,
+                          const std::vector<std::pair<int, int>>& pairs_edges, const fs::path& result_dir,
+                          bool save_single_runs = true, bool save_percentiles = true);
```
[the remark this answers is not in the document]
HenrZu: i see your point, otherwise one can definitely argue that there are also reason to leave both in the results_io ;) However, im also fine with moving this to the mobiliy_io header. 
its now shifted to the mobility header.

reneSchm (commented):
I think functionality wise this branch is fine, only the documentation needs a little touch up.

Could you please also add a short comment at the start of each new test that explains what it does, and, optimally, states the test case?

cpp/memilio/mobility/metapopulation_mobility_instant.h
```diff
@@ -138,6 +139,38 @@ class MobilityParameters
...
+     * @brief Constructor for initializing mobility parameters with coefficients from type MobilityCoefficientGroup and
+     * specific save indices.
+     *
+     * @param coeffs A group of mobility coefficients represented by a `MobilityCoefficientGroup` object, defining how 
+     * individuals move between nodes.
+     * @param save_indices A 2D vector of indices. Each inner vector represents a set of compartments whose data will be 
+     * saved in the member `m_mobility_results` of the `MobilityEdge` class during the simulation using the
+     * `add_mobility_result_time_point` function. 
+     */
+    MobilityParameters(const MobilityCoefficientGroup& coeffs, const std::vector<std::vector<size_t>>& save_indices)
```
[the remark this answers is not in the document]
reneSchm: Still missing in e.g. "get_save_indices". Please double check other locations.

cpp/memilio/mobility/metapopulation_mobility_instant.h
```diff
@@ -271,6 +321,38 @@ class MobilityEdge
...
+        , m_mobility_results(m_saved_compartment_indices.size() + 1)
+    {
+    }
+
+    /**
+     * create edge with coefficients as MobilityParameters object and a 2d vector of indices which determine which compartments we save.
+     * @param coeffs % of people in each group and compartment that migrate in each time step.
+     * @param save_indices 2D vector of indices. Each inner vector represents a group of indices to be saved.
+     */
+    MobilityEdge(const MobilityParameters<FP>& params, const std::vector<std::vector<size_t>>& save_indices)
```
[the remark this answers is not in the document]
reneSchm: Please also update save_edges (both overloads).

PR #994, opened 2024-04-05
305 add ccache to linux build

### LINKED ISSUE #305: Explore using ccache for CI pipeline
ccache could reduce the load and runtime of our pipeline a lot, since most files don't change in a pull request. How much can be gained needs to be tested, since use of templates makes ccache less efficient because there are fewer compilation units overall and changes are more likely.

dabele:
# Changes and Information

- added ccache to CI linux builds (other builds maybe added later after some observation)
- simplified linux build options, removed unused options to make the cache key easier

To see that the cache is working correctly: 
- The ccache action reports loading the cache at the beginning of the build job and cache misses/hits at the end
- stored cache files can be viewed in github -> actions -> caches; this development branch doesn't save caches anymore but there are some cache files around from my testing
- I tested the cache on the development branch before I added the restriction to only save on the main branch. compile times went down from ~15 mins to ~5 mins. this is with identical code, i.e. 100% cache hits. Compile times should be checked after a few nightly builds to see if everything works there.

Closes #305 

### Checks by code author


mknaranja:
@dabele Who has been involved here and who is supposed to do the review?

dabele:
> @dabele Who has been involved here and who is supposed to do the review?

it's not done yet. but I need the CI to run for testing this, so it's not a draft.

.github/actions/linux-build/action.yml
```diff
@@ -71,6 +55,21 @@ runs:
...
+    - name: ccache 
+      id: ccache
+      uses: hendrikmuhs/ccache-action@v1.2
+      with:
+        # free total space for cache available on github is limited to 10GB, so only save on main branch;
+        # other branches will try to use cache from main, but branches that are too old won't benefit;
+        save: ${{ github.ref_name == 'main' }} 
+        # set enough space per build cache to keep a few previous versions to be used by older branches
+        # debug builds require more space
+        max-size: ${{ inputs.config == 'Release' && '200M' || '1G' }}
```
reneSchm, line 67: Where do these sizes come from? More importantly, is there some report or warning if or when these sizes become too small?

.github/actions/linux-build/action.yml
```diff
@@ -71,6 +55,21 @@ runs:
...
+      with:
+        # free total space for cache available on github is limited to 10GB, so only save on main branch;
+        # other branches will try to use cache from main, but branches that are too old won't benefit;
+        save: ${{ github.ref_name == 'main' }} 
+        # set enough space per build cache to keep a few previous versions to be used by older branches
+        # debug builds require more space
+        max-size: ${{ inputs.config == 'Release' && '200M' || '1G' }} 
+        # saves cache as `ccache-<key>-<timestamp>` (timestamp added automatically, key must be unique, github caches are never overwritten)
+        # key should be a composite of all options that affect the compilation
+        key: linux-${{ inputs.compiler }}-${{ inputs.version }}-${{ inputs.config }}-cov${{ inputs.coverage }}-dep${{ inputs.optional-dependencies }}-omp${{ inputs.openmp }}-san${{ inputs.sanitizers }}
```
reneSchm, line 70: Can you add a small comment at the start of 'inputs:' as a reminder to change/update this key when any changes to the input are made?

Do we need to manage old caches after a change of keys, or are they deleted automatically at some point?

.github/actions/linux-build/action.yml
```diff
@@ -96,7 +95,7 @@ runs:
           exit 1
         fi
         mkdir -p build && cd build
-        cmake -DCMAKE_BUILD_TYPE=${{ inputs.config }} -DMEMILIO_BUILD_TESTS=${{ inputs.build-tests }} -DMEMILIO_BUILD_BENCHMARKS=${{ inputs.build-benchmarks }} -DMEMILIO_ENABLE_IPOPT=${{ inputs.enable-optimization }} -DMEMILIO_TEST_COVERAGE=${{ inputs.coverage }} -DMEMILIO_SANITIZE_ADDRESS=${{ inputs.sanitize-addr }} -DMEMILIO_SANITIZE_UNDEFINED=${{ inputs.sanitize-ub }} -DMEMILIO_USE_BUNDLED_JSONCPP=${{ inputs.optional-dependencies }} -DMEMILIO_ENABLE_OPENMP=${{ inputs.openmp }} ..
+        cmake -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ inputs.config }} -DMEMILIO_ENABLE_IPOPT=ON -DMEMILIO_TEST_COVERAGE=${{ inputs.coverage }} -DMEMILIO_SANITIZE_ADDRESS=${{ inputs.sanitizers }} -DMEMILIO_SANITIZE_UNDEFINED=${{ inputs.sanitizers }} -DMEMILIO_USE_BUNDLED_JSONCPP=${{ inputs.optional-dependencies }} -DMEMILIO_ENABLE_OPENMP=${{ inputs.openmp }} ..
```
reneSchm, line 98:
```suggestion
        cmake -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ inputs.config }} -DMEMILIO_BUILD_BENCHMARKS=ON -DMEMILIO_ENABLE_IPOPT=ON -DMEMILIO_TEST_COVERAGE=${{ inputs.coverage }} -DMEMILIO_SANITIZE_ADDRESS=${{ inputs.sanitizers }} -DMEMILIO_SANITIZE_UNDEFINED=${{ inputs.sanitizers }} -DMEMILIO_USE_BUNDLED_JSONCPP=${{ inputs.optional-dependencies }} -DMEMILIO_ENABLE_OPENMP=${{ inputs.openmp }} ..
```
I think we should still build the benchmarks, they are often enough neglected when updating related code, without the CI this may not be noticed before a merge.

reneSchm (changes requested):
This looks very nice, I do have a few questions (and one suggestion) though.

A little off topic, but does this already help with the boost dependency, or do we need another solution for that? It currently takes up a lot of CI time, both for the download itself as well as for packing and transferring artifacts, as these pack the entire build directory.

.github/actions/linux-build/action.yml
dabele: basically, I looked at the size of the cache for one build and at least doubled it. debug builds require 300-400M for a single build. So there is room for 2-3 older versions. release builds are much smaller, <50M, so there is more room here. It's possible we want to tune those numbers later, give debug builds a bit more space. I have little data on how fast this will grow with the code changes we are making. 

It's important to stay below the 10G limit, otherwise caches will be evicted immediately after saving. Currently, there are 10 builds cached, so 1G on average per build is the absolute limit. 

I wanted to save a bit of room for caching other things, e.g. boost or other dependencies.

.github/actions/linux-build/action.yml
dabele: good idea.

we don't need to manage caches, when the limit of 10G is reached, old caches are evicted automatically. New caches are never lost.

.github/actions/linux-build/action.yml
dabele: probably true, but should be a separate issue.

dabele:
> A little off topic, but does this already help with the boost dependency, or do we need another solution for that? It currently takes up a lot of CI time, both for the download itself as well as for packing and transferring artifacts, as these pack the entire build directory.

Sadly, this doesn't help with that. I though about caching boost, but this would only be possible if the same bootstrapped boost is usable by all builds (or at least all builds of the same operating system family). boost is more than 1G, so if every build needs it's own cached version this will exceed the limit. This will only help with the download and bootstrap phase of boost, not sure what could be done about the artifact size. I guess only saving artifacts that are really required, i.e. test executables and shared objects.

.github/actions/linux-build/action.yml
reneSchm: That sounds reasonable. Maybe you could mention the 10G limit (or link the github page that mentions it) in the comment here.

The cache size is shown, but only under "Post Run" in a collapsed line "cache stats". I think it might be useful to make this more visible, there is a format that turns echo into an [annotation](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-creating-an-annotation-for-an-error). An error if the cache is larger than expected would be nice.

.github/actions/linux-build/action.yml
reneSchm: Oh right, the default for "inputs.build-benchmarks" was "off", and it seems like it was never set. Opened #1045

.github/actions/linux-build/action.yml
dabele: you can also see the cache sizes in the github web interface in Actions -> Cache.

The question is, how large is too large? The cache will never be bigger than the limit (e.g. 200M for release builds), because of automatic eviction. But after some code changes, the cache will certainly hit the limit and then never shrink again. The cache will remain around the limit. So only the cache hits and misses can give more information. But I don't know how many hits or misses are expected, because it depends on the changes in the code. One extreme case would be a less than 100% hit rate where there have been no changes since the last build. But that would be complicated to implement, comparing github log and time stamps of the cache, etc. Not convinced this is worth it. Maybe with more data there can be a heuristic.

dabele: not sure what you mean. the size of the ccache directory will just be the limit set in the yaml (after some code changes of course)

reneSchm: The idea is to try and look whether ccache tries to use more space than the action allows for. But if ccache limits its output/cache on its own, that wont work.

PR #996, opened 2024-04-09
988 info on sucessful adaptive step sizing is also shown for fixed step size integrators

### LINKED ISSUE #988: Info on sucessfull adaptive step sizing is also shown for fixed-step-size integrators

If using an EulerIntegratorCore and LogLevel::info, adaptive step sizing is reported to be successful although there was no adaptivity.

### Enhancement description

The message should be replaced or better removed for fixed-step-size integrators.

### Additional context

_No response_

### Checklist

- [X] Linked to project

reneSchm:
# Changes and Information

- Check whether an IntegratorCore is adaptive in OdeIntegrator::advance.
- Fix floating_point_less_equal.

- Checks only whether dt changes during m_core->step calls. After only one change, register m_core as adaptive.
- The result "m_is_adaptive" is stored in OdeIntegrator, and it is only reset (to false) whenever m_core is set.
  - This may prevent under-reporting, e.g. when advance is called repeatedly with `dt == dt_max` while the ODE admits a larger step size.
  - While currently not possible, this may cause over-reporting if an adaptive m_core were to be configured to be non-adaptive in between advance calls. I think this is acceptable.
- I changed the while loop to a for loop to have the loop statement in one place. This only changes the scope of `i`. I you prefer the while loop for any reason, I will undo this change.
- I spotted that `floating_point_less_equal` returned  `!floating_point_less` and corrected it to `!floating_point_greater`.

### Checks by code author

Closes #988

### DIFF
--- a/cpp/memilio/math/floating_point.h
+++ b/cpp/memilio/math/floating_point.h
@@ -110,7 +110,7 @@ bool floating_point_greater(T v1, T v2, T abs_tol = 0, T rel_tol = std::numeric_
 template <class T>
 bool floating_point_less_equal(T v1, T v2, T abs_tol = 0, T rel_tol = std::numeric_limits<T>::min())
 {
-    return !floating_point_less(v2, v1, abs_tol, rel_tol);
+    return !floating_point_greater(v2, v1, abs_tol, rel_tol);
 }
 
 /**
--- a/cpp/memilio/math/integrator.cpp
+++ b/cpp/memilio/math/integrator.cpp
@@ -38,10 +38,11 @@ Eigen::Ref<Eigen::VectorXd> OdeIntegrator::advance(const DerivFunction& f, const
 
     bool step_okay = true;
 
+    double dt_cp; // used to check whether step sizing is adaptive
     double dt_restore = 0; // used to restore dt if dt was decreased to reach tmax
     double t          = t0;
-    size_t i          = results.get_num_time_points() - 1;
-    while (std::abs((tmax - t) / (tmax - t0)) > 1e-10) {
+
+    for (size_t i = results.get_num_time_points() - 1; std::abs((tmax - t) / (tmax - t0)) > 1e-10; ++i) {
         //we don't make timesteps too small as the error estimator of an adaptive integrator
         //may not be able to handle it. this is very conservative and maybe unnecessary,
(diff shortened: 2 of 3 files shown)


cpp/memilio/math/floating_point.h
```diff
@@ -72,7 +72,7 @@ template <class T>
 bool floating_point_less(T v1, T v2, T abs_tol = 0, T rel_tol = std::numeric_limits<T>::min())
 {
     auto diff = v1 - v2;
-    return diff < -abs_tol || diff < -abs_max(v1, v2) * rel_tol;
+    return diff < abs_tol || diff < abs_max(v1, v2) * rel_tol;
```
mknaranja, line 75: I think the change is not correct as now `a) v1 not == v2 within tolerances` is not satisfied anymore.

reneSchm: You are right, I misunderstood the function. But this means that increasing a tolerance *reduces* how "tolerant" the function is, correct? I think this is misleading.

Moreover, I think the old version does not work correctly as well: Take `v1=1.1, v2=3.2, abs_tol=20 , rel_tol=0`. Then we have
```
floating_point_less(v1, v2, abs_tol, 0.0) == true;
```
since ` -2.1 < -0` from the rel_tol comparision holds true, but the values are certainly equal w.r.t abs_tol. Is this as intended?

reneSchm: I think the correct version might be:
```cpp
bool floating_point_less(T v1, T v2, T abs_tol = 0, T rel_tol = std::numeric_limits<T>::min())
{
    auto diff = v1 - v2;
    return diff < -(abs_tol + rel_tol * abs_max(v1, v2));
}
```
I already pushed it, and in the tests there is a truth table that shows how this behaves. Could you please double check that this has the intended behavior?

reneSchm: At least with this version, `floating_point_less_equal == !floating_point_greater` and `floating_point_greater_equal == !floating_point_less` hold.

mknaranja: perfect

mknaranja (approved):

mknaranja:
@reneSchm Thanks for the nice fix and the additional work on float comparison operators. Good job.

PR #1035, opened 2024-05-15
1034 ide model simulation does not work under some conditions

### LINKED ISSUE #1034: IDE model simulation does not work under some conditions

Currently, we have that in the function compute_compartment_from_flows(), the idx_TransitionDistribution2 is set to 0 if not stated otherwise. On the other hand, it is not necessary nor specified in the check_constraints() function that the TransitionDistribution for SusceptibleToExposed is a StateAgeFunction of Type b).
For example, if you use ConstantFunction for TransitionDistribution[0], the simulation will not work properly because the maximum support of the ConstantFunction is used. 

### Enhancement description

A quick fix should be to include an if clause for the calculation of calc_time in compute_compartment_from_flows().

### Additional context

_No response_

### Checklist

- [X] Linked to project

lenaploetzke:
# Changes and Information

- compute_compartment_from_flows() now works when the TransitionDistribution SusceptibleToExposed is set to a ConstantFunction (or any other function with an invalid maximum support).
- fixed missing space in error message

-
-

### Checks by code author

Closes #1034.

### DIFF
--- a/cpp/models/ide_secir/model.cpp
+++ b/cpp/models/ide_secir/model.cpp
@@ -60,12 +60,17 @@ void Model::compute_compartment_from_flows(ScalarType dt, Eigen::Index idx_Infec
                                            Eigen::Index idx_IncomingFlow, int idx_TransitionDistribution1,
                                            int idx_TransitionDistribution2)
 {
-    ScalarType sum = 0;
-
+    ScalarType sum       = 0;
+    ScalarType calc_time = 0;
     // Determine relevant calculation area and corresponding index.
-    ScalarType calc_time =
-        std::max(parameters.get<TransitionDistributions>()[idx_TransitionDistribution1].get_support_max(dt, m_tol),
-                 parameters.get<TransitionDistributions>()[idx_TransitionDistribution2].get_support_max(dt, m_tol));
+    if ((1 - parameters.get<TransitionProbabilities>()[idx_TransitionDistribution1]) > 0) {
+        calc_time =
+            std::max(parameters.get<TransitionDistributions>()[idx_TransitionDistribution1].get_support_max(dt, m_tol),
+                     parameters.get<TransitionDistributions>()[idx_TransitionDistribution2].get_support_max(dt, m_tol));
+    }
+    else {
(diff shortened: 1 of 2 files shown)


cpp/models/ide_secir/model.cpp
```diff
@@ -60,12 +60,17 @@ void Model::compute_compartment_from_flows(ScalarType dt, Eigen::Index idx_Infec
...
 {
-    ScalarType sum = 0;
-
+    ScalarType sum       = 0;
+    ScalarType calc_time = 0;
     // Determine relevant calculation area and corresponding index.
-    ScalarType calc_time =
-        std::max(parameters.get<TransitionDistributions>()[idx_TransitionDistribution1].get_support_max(dt, m_tol),
-                 parameters.get<TransitionDistributions>()[idx_TransitionDistribution2].get_support_max(dt, m_tol));
+    if ((1 - parameters.get<TransitionProbabilities>()[idx_TransitionDistribution1]) > 0) {
```
annawendler, line 66:
```suggestion
    if ((parameters.get<TransitionProbabilities>()[idx_TransitionDistribution2]) > 0) {
```
I find that a bit more intuitive but don't have a strong opinion here.

lenaploetzke: That is actually wrong because idx_TransitionDistribution2 can be a random index if the second distribution is not used (eg for E->C). This is the whole problem of this issue/PR. 

I used 1- prob[idx]>0 because this is the factor used in the calculation below, another opportunity would of course be prob[idx1]<1.

annawendler: But if the second TransitionDistribution is not used the default index is 0 and for this the probability is 1. But maybe I'm confused here.

But it does make sense to use the first Transition Distribution in the if statement because this already determines if we need both distributions or not, so it is fine as it is.

annawendler (approved):

cpp/models/ide_secir/model.cpp
lenaploetzke: Yes you are right, we actually have a test that ensures that this probability is one. :) I still prefer the other solution.

PR #1046, opened 2024-05-30
1038 odeintegrator can run indefinitely

### LINKED ISSUE #1038: OdeIntegrator can run indefinitely
### Bug description

After fixing a previous issue, where the IntegratorCore made steps smaller than dt_min, the OdeIntegrator can now no longer converge in certain situations, i.e. if `t + dt_min > t_max` without reaching t_max within tolerances, then dt will be set to a negative value by the OdeIntegrator and reset to dt_min by the Core, indefinitely. 

The problem lies in the requirement that the Core may not make steps smaller than dt_min. This needs to be changed or circumvented.

### Version

Any

### To reproduce

Set e.g. dt=dt_min=0.7, t_0=0, t_max=1 for any integration/simulation and run it.

### Relevant log output

_No response_

_No response_

### Checklist

- [X] Linked to project

reneSchm:
# Changes and Information

- Fix a bug where OdeIntegrator would run indefinitely.
- Allow integrators to force a given step size.

- Please look through the newly added tests. They impose assumptions on the IntegratorCore::step method we may want to discuss.
  - e.g., a forced step currently always counts as successful 
  - is there maybe something missing?
- Please proofread the documentation, especially for IntegratorCore::step
- Should we document the unused argument force_step_size of EulerIntegratorCore::step? How?

### Checks by code author

Closes #1038


cpp/memilio/math/euler.h
```diff
@@ -42,8 +42,8 @@ class EulerIntegratorCore : public IntegratorCore<FP>
      * @param[in,out] dt current time step h=dt
      * @param[out] ytp1 approximated value y(t+1)
```
lenaploetzke, line 43: Something like:
* @param The last argument is not used, as the Euler method is not adaptive.
?

cpp/tests/test_graph_simulation.cpp
```diff
@@ -237,9 +237,9 @@ TEST(TestGraphSimulation, consistencyStochasticMobility)
     auto result_n0 = sim.get_graph().nodes()[0].property.get_result().get_last_value();
     auto result_n1 = sim.get_graph().nodes()[1].property.get_result().get_last_value();
 
-    auto expected_values_n0 = std::vector<double>{692.0, 43.630772796677256, 95.750528156188381, 159.61869904713436};
+    auto expected_values_n0 = std::vector<double>{692.0, 43.647065696491225, 95.758901343443526, 159.59403296006528};
     auto actual_values_n0   = std::vector<double>{result_n0[0], result_n0[1], result_n0[2], result_n0[3]};
-    auto expected_values_n1 = std::vector<double>{708.0, 44.063147085799322, 96.485223892060375, 160.45162902214025};
+    auto expected_values_n1 = std::vector<double>{708.0, 44.04685402850226, 96.476851270558342, 160.47629470093941};
```
lenaploetzke, line 242: Can you explain to me why these values have to be changed?

cpp/memilio/math/euler.h
reneSchm: This would lead to a parameter named "The". But we could name it and mark it as unused in the function body?

cpp/tests/test_graph_simulation.cpp
reneSchm: Yes, it is due to the newly forced steps. Disabling forced step sizing (i.e. overriding the bool with true) does recover the old results.
The graph simulation makes small time steps in which it calls each node simulation, which usually calls advance with `dt < t_max - t`, hence it makes only one forced time step. There are at least two steps smaller than dt_min, hence the discrepancy.

reneSchm: Actually, there was a problem with the OdeIntegrator. In its advance function, I forced a step whenever making the last step, i.e. `dt < t_max - t`. However, we do not care whether this is the last step or not, we want to know whether the last step would be smaller than dt_min (so whenever the step with `dt = t_max - t` would fail to adapt). Only then we actually need to force a smaller step, otherwise we may loose accuracy.

The new logic in advance re-takes a step, which is not optimal, but currently there is no way to get dt_min from advance. Maybe we need another virtual method?

lenaploetzke:
With these changes, the application where the problem occurred works as expected. Will review the code soon.

cpp/memilio/math/euler.h
```diff
@@ -41,10 +42,12 @@ class EulerIntegratorCore : public IntegratorCore<FP>
...
+     * @param[in] force_step_size This argument is unused, as the step width is always fixed.
      */
-    bool step(const DerivFunction<FP>& f, Eigen::Ref<const Vector<FP>> yt, FP& t, FP& dt,
-              Eigen::Ref<Vector<FP>> ytp1) const override
+    bool step(const DerivFunction<FP>& f, Eigen::Ref<const Vector<FP>> yt, FP& t, FP& dt, Eigen::Ref<Vector<FP>> ytp1,
+              const bool force_step_size = false) const override
     {
+        mio::unused(force_step_size);
```
lenaploetzke, line 50: I like this solution.

cpp/memilio/math/integrator.h
```diff
@@ -60,11 +60,14 @@ class IntegratorCore
      * @param[out] ytp1 Set to the approximated value of y at time t + dt.
      *     (If adaptive, this time may be smaller, but it is at least t + dt_min, at most t + dt_max.
```
lenaploetzke, line 61: Should this be adjusted?

cpp/memilio/math/integrator.h
```diff
@@ -127,19 +130,30 @@ class OdeIntegrator
...
+            // catch case where dt_min of an adaptive stepper is larger than tmax - t
+            if (t > tmax) {
+                log_info("Forcing last step size.");
+                // reset time and step size
+                dt = dt_copy;
+                t  = tmax - dt_copy;
+                // recalculate the erroneous last step and overwrite its results
+                step_okay = m_core->step(f, results[i], t, dt, results[i + 1], true);
+            }
+            is_step_sizing_okay &= step_okay;
```
lenaploetzke, line 147: I think it would be enough to just use one bool.

cpp/memilio/math/stepper_wrapper.h
```diff
@@ -70,24 +72,27 @@ class ControlledStepperWrapper : public mio::IntegratorCore<FP>
...
-    * @param[in,out] dt Current time step size h=dt. Overwritten by an estimated optimal step size for the next step.
-    * @param[out] ytp1 The approximated value of y(t_{k+1}).
-    */
+     * @brief Make a single integration step on a system of ODEs and adapt the step size dt.
+     *
+     * @param[in] yt Value of y at t_{k}, y(t_{k}).
+     * @param[in,out] t Current time step t_{k} for some k. Will be set to t_{k+1} in [t_{k} + dt_min, t_{k} + dt].
+     * @param[in,out] dt Current time step size h=dt. Overwritten by an estimated optimal step size for the next step.
+     * @param[out] ytp1 The approximated value of y(t_{k+1}).
+     * @param[in] force_step_size Forces a step with exactly size dt.
```
lenaploetzke, line 81: Maybe a little more detailed (as in the integrator)?

lenaploetzke (commented):
Thank you for the work! I still need to review the test.

reneSchm:
> Nevertheless, I still get the warning "Adaptive step sizing failed. Forced at least one integration step of size dt_min." if the last step is forced.

The message should only appear if a non-forced step failed to adapt, i.e. the stepper wanted to make a step smaller than dt_min. Are you sure this message is due to the last step?

cpp/memilio/math/integrator.h
reneSchm: I decided against it to not duplicate information, but I think I could just add "If adaptive and force_step_size==false" or something similar to hint at the other parameter.

cpp/memilio/math/integrator.h
reneSchm: With only one bool and a forced last step it is difficult to report what the state before the forced step was. I assume that if we discard the step that went beyond tmax, we also want to discard its "step_okay" value. And if that step was not okay (i.e. false), the &= will overwrite the previous state, making it unrecoverable.

Do you agree with this assumption?

HenrZu (changes requested):
Just a few short comments. Otherwise really great, especially the tests :)

cpp/memilio/math/adapt_rk.h
```diff
@@ -208,19 +208,22 @@ class RKIntegratorCore : public IntegratorCore<FP>
...
      * @param[in,out] dt current time step size h=dt
      * @param[out] ytp1 approximated value y(t+1)
+     * @param[in] force_step_size Forces a step with exactly size dt.
      */
     bool step(const DerivFunction<FP>& f, Eigen::Ref<Eigen::VectorXd const> yt, double& t, double& dt,
-              Eigen::Ref<Eigen::VectorXd> ytp1) const override
+              Eigen::Ref<Eigen::VectorXd> ytp1, const bool force_step_size = false) const override
     {
         assert(0 <= m_dt_min);
         assert(m_dt_min <= m_dt_max);
```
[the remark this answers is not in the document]
reneSchm: IMO this is a perfect use case for for an assert, since we *cannot* continue if these conditions fail, but checking this every iteration is needlessly expensive.
And since the asserts are printed in plain text on failure, I don't see the need to add text to it.

cpp/memilio/math/integrator.h
```diff
@@ -127,19 +130,30 @@ class OdeIntegrator
...
             results.add_time_point();
-            step_okay &= m_core->step(f, results[i], t, dt, results[i + 1]);
-            results.get_last_time() = t;
+            bool step_okay = m_core->step(f, results[i], t, dt, results[i + 1]);
 
             // if dt has been changed (even slighly) by step, register the current m_core as adaptive
             m_is_adaptive |= !floating_point_equal(dt, dt_copy);
+
+            // catch case where dt_min of an adaptive stepper is larger than tmax - t
+            if (t > tmax) {
```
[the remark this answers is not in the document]
reneSchm: We do have floating_point compare functions, which work at least with those ad types, and can use both relative and absolute tolerances. To be clear, that tolerance is then used as "t >= tmax, and w.r.t. tolerances t != tmax" ([see here](https://github.com/SciCompMod/memilio/blob/main/cpp/memilio/math/floating_point.h#L76))

Do you think that tolerance is needed? Should we use the same tolerance as in the for loop?

cpp/memilio/math/euler.h
```diff
@@ -22,6 +22,7 @@
 
 #include "memilio/config.h"
 #include "memilio/math/integrator.h"
+#include "memilio/utils/compiler_diagnostics.h"
```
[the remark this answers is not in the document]
reneSchm: It provides mio::unused. You can check this via clangd.

cpp/memilio/math/stepper_wrapper.h
```diff
@@ -70,24 +72,27 @@ class ControlledStepperWrapper : public mio::IntegratorCore<FP>
...
         using std::max;
         assert(0 <= m_dt_min);
         assert(m_dt_min <= m_dt_max);
 
-        if (dt < m_dt_min || dt > m_dt_max) {
+        const FP dt_forced = dt;
+        if ((dt < m_dt_min || dt > m_dt_max) && !force_step_size) {
             mio::log_warning("IntegratorCore: Restricting given step size dt = {} to [{}, {}].", dt, m_dt_min,
                              m_dt_max);
+            dt = std::min(dt, m_dt_max);
```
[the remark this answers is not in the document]
reneSchm: I do not use clamp here nor in adapt_rk as a small optimization, as the main loop of both integrators already check and bound to dt_min.

Though the std::min should trip up some AD simulation, don't the tests cover simulations?

cpp/memilio/math/stepper_wrapper.h
```diff
@@ -115,18 +120,18 @@ class ControlledStepperWrapper : public mio::IntegratorCore<FP>
...
         // output the new value by copying it back to the reference
         ytp1 = m_ytp1;
-        // bound dt from below
-        // the last adaptive step (successful or not) may have calculated a new step size smaller than m_dt_min
-
-        dt = max(dt, m_dt_min);
+        // clamp dt (again)
+        // the last adaptive step (successful or not) may have calculated a new step size smaller than m_dt_min,
+        // and when forcing step size, dt may have been larger than m_dt_max
+        dt = std::clamp(dt, m_dt_min, m_dt_max);
```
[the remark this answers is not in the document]
reneSchm: No, as mentioned in the comment. Though it is now technically possible for a successful step without forcing to grow larger than dt_max, I will update that.

I also cannot disable the adaptive step sizing (without some additional code or run time cost).

cpp/memilio/math/adapt_rk.h
```diff
@@ -208,19 +208,22 @@ class RKIntegratorCore : public IntegratorCore<FP>
...
-                             m_dt_max);
-        }
+        if (!force_step_size) {
+            if (dt < m_dt_min || dt > m_dt_max) {
+                mio::log_warning("IntegratorCore: Restricting given step size dt = {} to [{}, {}].", dt, m_dt_min,
+                                 m_dt_max);
+            }
 
-        dt = std::min(dt, m_dt_max);
+            dt = std::min(dt, m_dt_max);
```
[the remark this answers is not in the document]
reneSchm: > Why is this line not part of the if statement?

I thought about it, and the code behavior is not affected by the placement of the min. Personally, I like it more outside such that the if only affects the log message, and no further code behavior. I can also move the min down above the while loop so it is closer to the "max", i.e. the if statement right at the beginning of the while. Maybe this also avoids confusion about the missing max/clamp?

reneSchm:
Closed in favor of v2.

PR #1049, opened 2024-06-11
1038 OdeIntegrator can run indefinitely v2

reneSchm:
# Changes and Information

- Fix a bug where OdeIntegrator would run indefinitely.
- Add dt_min/dt_max to IntegratorCore.

-
-

### Checks by code author

### DIFF
--- a/cpp/memilio/math/adapt_rk.h
+++ b/cpp/memilio/math/adapt_rk.h
@@ -192,7 +192,7 @@ public:
         this->get_dt_max() = dt_max;
     }
 
-    // Allow setting different RK tablea schemes
+    // Allow setting different RK tableau schemes
     void set_tableaus(const Tableau& tab, const TableauFinal& final_tab)
     {
         m_tab       = tab;
--- a/cpp/memilio/math/adapt_rk.h
+++ b/cpp/memilio/math/adapt_rk.h
@@ -148,10 +148,9 @@ public:
      * @brief Setting up the integrator
      */
     RKIntegratorCore()
-        : m_abs_tol(1e-10)
+        : IntegratorCore<FP>(std::numeric_limits<double>::min(), std::numeric_limits<double>::max())
+        , m_abs_tol(1e-10)
         , m_rel_tol(1e-5)
-        , m_dt_min(std::numeric_limits<double>::min())
-        , m_dt_max(std::numeric_limits<double>::max())
     {
     }
 
@@ -163,10 +162,9 @@ public:
      * @param dt_max upper bound for time step dt
      */
     RKIntegratorCore(const double abs_tol, const double rel_tol, const double dt_min, const double dt_max)
-        : m_abs_tol(abs_tol)
+        : IntegratorCore<FP>(dt_min, dt_max)
+        , m_abs_tol(abs_tol)
         , m_rel_tol(rel_tol)
-        , m_dt_min(dt_min)
-        , m_dt_max(dt_max)
     {
     }
 
@@ -185,13 +183,13 @@ public:
     /// @param dt_min sets the minimum step size
     void set_dt_min(double dt_min)
     {
-        m_dt_min = dt_min;
+        this->get_dt_min() = dt_min;
     }
 
     /// @param dt_max sets the maximum step size
     void set_dt_max(double dt_max)
     {
-        m_dt_max = dt_max;
+        this->get_dt_max() = dt_max;
     }
 
     // Allow setting different RK tablea schemes
@@ -212,15 +210,15 @@ public:
     bool step(const DerivFunction<FP>& f, Eigen::Ref<Eigen::VectorXd const> yt, double& t, double& dt,
               Eigen::Ref<Eigen::VectorXd> ytp1) const override
     {
-        assert(0 <= m_dt_min);
-        assert(m_dt_min <= m_dt_max);
+        assert(0 <= this->get_dt_min());
+        assert(this->get_dt_min() <= this->get_dt_max());
 
-        if (dt < m_dt_min || dt > m_dt_max) {
-            mio::log_warning("IntegratorCore: Restricting given step size dt = {} to [{}, {}].", dt, m_dt_min,
-                             m_dt_max);
+        if (dt < this->get_dt_min() || dt > this->get_dt_max()) {
+            mio::log_warning("IntegratorCore: Restricting given step size dt = {} to [{}, {}].", dt, this->get_dt_min(),
+                             this->get_dt_max());
         }
 
-        dt = std::min(dt, m_dt_max);
+        dt = std::min(dt, this->get_dt_max());
 
         double t_eval; // shifted time for evaluating yt
         double dt_new; // updated dt
@@ -236,9 +234,9 @@ public:
         m_yt_eval = yt;
 
         while (!converged && !dt_is_invalid) {
-            if (dt < m_dt_min) {
+            if (dt < this->get_dt_min()) {
                 dt_is_invalid = true;
-                dt            = m_dt_min;
+                dt            = this->get_dt_min();
             }
             // compute first column of kt, i.e. kt_0 for each y in yt_eval
             f(m_yt_eval, t, m_kt_values.col(0));
@@ -281,9 +279,9 @@ public:
             // and to avoid dt_new -> dt for step decreases when |error_estimate - eps| -> 0
             dt_new *= 0.9;
             // check if updated dt stays within desired bounds and update dt for next step
-            dt = std::min(dt_new, m_dt_max);
+            dt = std::min(dt_new, this->get_dt_max());
         }
-        dt = std::max(dt, m_dt_min);
+        dt = std::max(dt, this->get_dt_min());
         // return 'converged' in favor of '!dt_is_invalid', as these values only differ if step sizing failed,
         // but the step with size dt_min was accepted.
         return converged;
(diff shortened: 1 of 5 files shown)


cpp/memilio/math/integrator.h
```diff
@@ -126,13 +177,21 @@ class OdeIntegrator
             }
             dt_copy = dt;
 
+            if (tmax - t < dt_min_copy) {
+                // reduce minimal step size low enough such that we do not step past tmax
+                // m_core->get_dt_min() = min((tmax - t0) * 1e-10, dt_min_copy);
```
HenrZu, line 182:
```suggestion
```

HenrZu (changes requested):
First feedback. Need some more time to check/ verify everything

cpp/memilio/math/integrator.h
```diff
@@ -112,8 +162,9 @@ class OdeIntegrator
         bool step_okay = true;
 
         FP dt_copy; // used to check whether step sizing is adaptive
-        FP dt_restore = 0; // used to restore dt if dt was decreased to reach tmax
-        FP t          = t0;
+        FP dt_restore  = 0.0; // used to restore dt if dt was decreased to reach tmax
```
lenaploetzke, line 165: Maybe adapt comment as dt_min_copy is also "used to restore dt if dt was decreased to reach tmax"?

cpp/memilio/math/adapt_rk.h
```diff
@@ -185,13 +183,13 @@ class RKIntegratorCore : public IntegratorCore<FP>
...
     }
 
     /// @param dt_max sets the maximum step size
     void set_dt_max(double dt_max)
     {
-        m_dt_max = dt_max;
+        this->get_dt_max() = dt_max;
     }
 
     // Allow setting different RK tablea schemes
```
lenaploetzke, line 195:
```suggestion
    // Allow setting different RK tableau schemes
```

cpp/tests/test_numericalIntegration.cpp
```diff
@@ -17,17 +17,15 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
+#include "actions.h"
 #include "memilio/math/euler.h"
 #include "memilio/math/adapt_rk.h"
 #include "memilio/math/stepper_wrapper.h"
```
lenaploetzke, line 23:
```suggestion
#include "memilio/math/stepper_wrapper.h"
#include "memilio/utils/logging.h"
```

lenaploetzke: Maybe add also the integrator.h adder as OdeIntegrator is used directly.

lenaploetzke (commented):
Nice work! I think this solution is much better than the first version. I only have a few small comments.

cpp/memilio/math/integrator.h
```diff
@@ -38,6 +39,19 @@ template <typename FP = double>
 class IntegratorCore
 {
 public:
+    /**
+     * @brief Initialize an IntegratorCore.
+     * Step size bounds are needed for adaptive integrators, see the step method for more detail.
+     * Fixed size steppers ignore those bounds and may use the default constructor for FP.
```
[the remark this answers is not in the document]
reneSchm: My idea was to nudge users to look at this documentation by not having a default constructor, and especially avoid `dt_min=dt_max=0` for adaptive steppers. We can set different defaults, but that potentially could also cause issues. And it is not that much more difficult to use.

I am not completely against adding a default ctor, but I think not having one might be better.

cpp/memilio/math/integrator.h
```diff
@@ -100,6 +149,7 @@ class OdeIntegrator
         // hint at std functions for ADL
         using std::fabs;
         using std::max;
+        using std::min;
```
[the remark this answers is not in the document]
reneSchm: It is not. It got removed resolving one of Henrik's comments.

cpp/memilio/math/integrator.h
reneSchm: Yes, though it restores dt_min. Should it be renamed to dt_min_restore as well? I think _copy is okay, and shorter.

cpp/memilio/math/integrator.h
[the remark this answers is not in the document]
lenaploetzke: When using non-adaptive methods, IMO it is better to have a default constructor, because dt_min and dt_max make no sense there.
Is there anything wrong with dt_min=0 and dt_max="FP_MAX"? In any case, you should read the documentation if you use the class..

cpp/memilio/math/integrator.h
lenaploetzke: Maybe rename dt_restore to dt_copy then? I think it makes sense to name variables that do the same thing similarly, just for dt and dt_min.

cpp/memilio/math/integrator.h
```diff
@@ -112,8 +162,9 @@ class OdeIntegrator
         bool step_okay = true;
 
         FP dt_copy; // used to check whether step sizing is adaptive
-        FP dt_restore = 0; // used to restore dt if dt was decreased to reach tmax
-        FP t          = t0;
+        FP dt_restore  = 0.0; // used to restore dt, if dt was decreased to reach tmax
+        FP dt_min_copy = m_core->get_dt_min(); // used to restore dt_min, if it was decreased to reach tmax
```
HenrZu, line 166: Since the comments are exactly the same, it is confusing that one is called restore and one is called copy.
If the comments remain the same, it should be called dt_min_restore

cpp/memilio/math/integrator.h
lenaploetzke: Ok. If you are happy with the current naming, leave it as it is :)

cpp/memilio/math/integrator.h
```diff
@@ -126,13 +177,20 @@ class OdeIntegrator
...
             step_okay &= m_core->step(f, results[i], t, dt, results[i + 1]);
             results.get_last_time() = t;
 
             // if dt has been changed (even slighly) by step, register the current m_core as adaptive
             m_is_adaptive |= !floating_point_equal(dt, dt_copy);
         }
+        m_core->get_dt_min() = dt_min_copy; // reset dt_min
         // if dt was decreased to reach tmax in the last time iteration,
         // we restore it as it is now probably smaller than required for tolerances
         dt = max(dt, dt_restore);
```
HenrZu, line 196: By resetting/restoring, we get back a dt, which does not reflect the true step size in the last time step.

cpp/tests/test_numericalIntegration.cpp
```diff
@@ -342,3 +359,33 @@ TEST(TestOdeIntegrator, integratorContinuesAtLastState)
...
+    };
+
+    const auto num_calls = Eigen::Index(t_max / dt_min) + 1;
+    EXPECT_CALL(*mock_core, step).Times(num_calls).WillRepeatedly(testing::Invoke(step_fct));
+
+    // run a mock integration to examine whether only the last step is forced
+    mio::TimeSeries<double> mock_result(0, Eigen::VectorXd::Constant(1, 0));
+    integrator.advance(f, t_max, dt, mock_result);
+
+    EXPECT_EQ(mock_result.get_num_time_points(), num_calls + 1);
```
HenrZu, line 387: as mentioned above, here `dt = 0.7`.

HenrZu (changes requested):

cpp/memilio/math/integrator.h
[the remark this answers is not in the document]
reneSchm: > you should read the documentation if you use the class..

Yeah, that would be a nice change...

> Is there anything wrong with dt_min=0 and dt_max="FP_MAX"?

Technically any choice can have problems. dt_min = 0 allows for infinite loops, and "FP_MIN" is not better on a human time scale. Any "larger" value may cause loss of precision.
But the main point is to move that choice away from the interface (IntegratorCore) and to the actual implementations. 

Maybe it would be better to split IntegratorCore into adaptive and non-adaptive? Then we could avoid the issue of "just ignore dt_min/_max" and simplify the documentation.

cpp/memilio/math/integrator.h
reneSchm: I don't think that is a problem. Adaptive steppers are expected to change dt anyways, and the "true" value is in the results.
The problem this line addresses is that, e.g., we have used dt~1e-3, but dt=tmax-t may have set dt~1e-8 or so. If we restart the integrator, this would cause several steps that are a lot smaller than needed. This is important for graph sims which pause integration frequently.

cpp/tests/test_numericalIntegration.cpp
reneSchm: But that is correct? Please explain.

cpp/memilio/math/integrator.h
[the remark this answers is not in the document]
lenaploetzke: I understand your arguments against default values. I dont like the idea of having two classes. It is an advantage, that we have just one class for all solvers ATM. Under these conditions, I think it's fine not to have a default constructor.

cpp/memilio/math/integrator.h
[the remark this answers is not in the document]
lenaploetzke: Did you add the line again?

lenaploetzke (approved):

HenrZu (approved):

PR #1068, opened 2024-07-11
932 Get_default of Parameter HighViralLoadProtectionFactor doesn't work as expected

### LINKED ISSUE #932: Get_default of Parameter HighViralLoadProtectionFactor doesn't work as expected
### Bug description

When you don't set HighViralLoadProtectionFactor it should default to a function which goves back 0.0 every time. This thrwos an error std::bad_function_call for me.

### Version

Any

### To reproduce

1. comment out any simulation part where HighViralLoadProtectionFactor is set
2. run simulation till the variable high_viral_load_factor is used in infection
3. error

### Relevant log output

```zsh
libc++abi: terminating due to uncaught exception of type std::__1::bad_function_call: std::exception
```

_No response_

### Checklist

- [X] Linked to project

khoanguyen-dev:
# Changes and Information

Revised Definition of HighViralLoadProtectionFactor:
- Change data type of `HighViralLoadProtectionFactor` to CustomIndexArray<InputFunctionForProtectionLevel, ExposureType, AgeGroup, VirusVariant>.
- Update python bindings to fix parameter (de)serialize.

-

### Checks by code author

- [X] Every addressed issue is linked (use the "Closes #ISSUE" keyword below)
- [X] New code adheres to [coding guidelines](https://github.com/SciCompMod/memilio/wiki/Coding-guidelines)
- [X] No large data files have been added (files should in sum not exceed 100 KB, avoid PDFs, Word docs, etc.)
- [X] Tests are added for new functionality and a local test run was successful (with and without OpenMP)
- [X] Appropriate **documentation** for new functionality has been added (Doxygen in the code and Markdown files if necessary)
- [X] Proper attention to licenses, especially no new third-party software with conflicting license has been added
- [X] (For ABM development) Checked [benchmark results](https://github.com/SciCompMod/memilio/wiki/Agent-Based-Model-Development) and ran and posted a local test above from before and after development to ensure performance is monitored.

---------------------------------------------------------------------------
Benchmark                                 Time             CPU   Iterations
---------------------------------------------------------------------------
abm_benchmark/abm_benchmark_50k        2721 ms         2716 ms            1
abm_benchmark/abm_benchmark_100k       5682 ms         5674 ms            1
abm_benchmark/abm_benchmark_200k      11855 ms        11773 ms            1

---------------------------------------------------------------------------
Benchmark                                 Time             CPU   Iterations
---------------------------------------------------------------------------
abm_benchmark/abm_benchmark_50k        3087 ms         2953 ms            1
abm_benchmark/abm_benchmark_100k       5698 ms         5690 ms            1
abm_benchmark/abm_benchmark_200k      11585 ms        11554 ms            1

Closes #932


DavidKerkmann:
Thanks for fixing this. It might also be relevant for `SeverityProtectionFactor` and `InfectionProtectionFactor`. Have you checked this?

khoanguyen-dev:
I have checked 

> Thanks for fixing this. It might also be relevant for `SeverityProtectionFactor` and `InfectionProtectionFactor`. Have you checked this?

I have checked them (see `TEST(TestInfection, getPersonalProtectiveFactor)`). I think they are initialised in `CustomIndexArray` so there is no need for a specified contructor.

DavidKerkmann:
Ah, yes, they have a different type. Why did we decide again that HighViralLoadProtectionFactor does not depend on age/virus type? Was this a result of some article?

khoanguyen-dev:
> Ah, yes, they have a different type. Why did we decide again that HighViralLoadProtectionFactor does not depend on age/virus type? Was this a result of some article?

Yes, I believe it is based on an article that mentions only about days since onset of symptoms (https://doi.org/10.1093/cid/ciaa886). I guess we could also adopt it to CustomIndexArray with a general age/virus type for consistency.

DavidKerkmann:
I understand our parameter `HighViralLoadProtectionFactor` as a protection to obtaining a high viral load when you have been vaccinated or previously infected. From what I can see from the article is only a plot of viral load over time and integral values of this curve which are set into relation with transmission potential, but there is no mention about any previous engagement of the immune system. Am I missing something or could you explain?

From a technical point of view, I agree with you that we should make this dependant on virus variant and age, too, as this might very well differ.

xsaschako (changes requested):
I think this is a little convoluted. I think we should just change it to AgeGroup as an input parameter and this should work. I think we did this for the paper branch.
Also i think for now just agegroup is fine, let's not make things more complicated as needed right now, and 2 viruses in a simulation (where this would come in handy) is not planned and i think we need to change a greater amount then either way.

DavidKerkmann:
> I unfortunatley think this is a little convoluted. I think one can just change it to AgeGroup /_Agegroup_/ as an input parameter and this should work. I think we did this for the paper branch.

In the paper branch we just initialised it explicitly.

> Also i think for now just agegroup is fine, let's not make things more complicated as needed right now, and 2 viruses in a simulation (where this would come in handy) is not planned and i think we need to change a greater amount then either way.

This is planned in the near future to study the overlap of two (or more) variants and to examine what it takes for one variant to dominate. I also don't think we have to change much in the rest of the codebase, as everything should already work with having two variants. Also, I don't think it created much of an overhead or making things more complicated. I leave the decision to you, @khoanguyen-dev .

xsaschako:
> This is planned in the near future to study the overlap of two (or more) variants and to examine what it takes for one variant to dominate.

Didn't know! Then it's good.

DavidKerkmann (approved):
We agreed to leave out a common initialization in the simulations for the HighViralLoadProtectionFactor, as there was no fitting literature. In the future, we might have to review this, perhaps in combination with the other ProtectionFactors.

DavidKerkmann:
Fixed the merge conflicts and fixed python bindings. (De)serialisation has to be postponed to #1072 .

DavidKerkmann:
Add a default constructor that initializes the function member to the result of get_default(). This ensures that the function is always properly initialized.
    An operator() is defined to allow instances of HighViralLoadProtectionFactor to be called like a function. This operator forwards the call to the function member.
    The function member is initialized in the constructor, ensuring that it always holds a valid callable.

This is from the changes on top of this PR. Is this still up-to-date? @khoanguyen-dev

khoanguyen-dev:
> ```
> Add a default constructor that initializes the function member to the result of get_default(). This ensures that the function is always properly initialized.
> An operator() is defined to allow instances of HighViralLoadProtectionFactor to be called like a function. This operator forwards the call to the function member.
> The function member is initialized in the constructor, ensuring that it always holds a valid callable.
> ```
> 
> This is from the changes on top of this PR. Is this still up-to-date? @khoanguyen-dev

I updated the description. It is now a simple extension for this parameter.

pycode/memilio-simulation/memilio/simulation/bindings/models/abm.cpp
```diff
@@ -117,26 +134,6 @@ PYBIND11_MODULE(_simulation_abm, m)
...
-    pymio::bind_CustomIndexArray<mio::abm::TestParameters, mio::abm::TestType>(m, "_TestData");
-    pymio::bind_Index<mio::abm::ExposureType>(m, "ExposureTypeIndex");
-    pymio::bind_ParameterSet<mio::abm::ParametersBase, pymio::EnablePickling::Never>(m, "ParametersBase");
-    pymio::bind_class<mio::abm::Parameters, pymio::EnablePickling::Never, mio::abm::ParametersBase>(m, "Parameters")
-        .def(py::init<int>())
-        .def("check_constraints", &mio::abm::Parameters::check_constraints);
-
-    pymio::bind_ParameterSet<mio::abm::LocalInfectionParameters, pymio::EnablePickling::Never>(
-        m, "LocalInfectionParameters")
-        .def(py::init<size_t>());
```
DavidKerkmann: Why was this removed again? I think we need it for the correct python bindings of the TestParameters, as the time and type are missing in the current bindings. Also, we wanted to include the serialization for the TestParameters once the auto-serialization has been done.

DavidKerkmann (commented):
I noticed that the pybinds were unchanged, perhaps this is a mistake from an earlier merge? I am confused why this was there in the first place, as it has nothing to do with this issue, but is also not on the main. What happened here?

pycode/memilio-simulation/memilio/simulation/bindings/models/abm.cpp
khoanguyen-dev: Sorry, I made a mistake during the manual merging with the main branch. I completely forgot that you corrected the python bindings of the TestParameters in the issue as well. It's now added back.

cpp/models/abm/parameters.h
```diff
@@ -296,13 +296,14 @@ struct SeverityProtectionFactor {
 };
 
 /**
- * @brief Personal protective factor against high viral load. Its value is between 0 and 1.
+ * @brief Personal protective factor against high viral load, , which depends on #ExposureType,
```
DavidKerkmann, line 299:
```suggestion
 * @brief Personal protective factor against high viral load, which depends on #ExposureType,
```

cpp/models/abm/parameters.h
```diff
@@ -296,13 +296,14 @@ struct SeverityProtectionFactor {
...
 /**
- * @brief Personal protective factor against high viral load. Its value is between 0 and 1.
+ * @brief Personal protective factor against high viral load, which depends on #ExposureType,
+ * #AgeGroup and #VirusVariant. Its value is between 0 and 1.
  */
 struct HighViralLoadProtectionFactor {
-    using Type = InputFunctionForProtectionLevel;
-    static auto get_default()
+    using Type = CustomIndexArray<InputFunctionForProtectionLevel, ExposureType, AgeGroup, VirusVariant>;
+    static auto get_default(AgeGroup size)
```
reneSchm, line 304:
```suggestion
    using Type = CustomIndexArray<TimeSeriesFunctor<ScalarType>, ExposureType, AgeGroup, VirusVariant>;
    static auto get_default(AgeGroup size)
```

cpp/models/abm/parameters.h
```diff
@@ -296,13 +296,14 @@ struct SeverityProtectionFactor {
...
     {
-        return Type([](ScalarType /*days*/) -> ScalarType {
+        return Type({ExposureType::Count, size, VirusVariant::Count}, [](ScalarType /*days*/) -> ScalarType {
             return 0;
         });
```
reneSchm, line 308:
```suggestion
        return Type({ExposureType::Count, size, VirusVariant::Count}, TimeSeriesFunctor<ScalarType>());
```

As a heads up, the InputFunctionForProtectionLevel aka std::function was deleted in the serialization branch, so this has to be updated.

pycode/memilio-simulation/memilio/simulation/bindings/models/abm.cpp
```diff
@@ -134,6 +117,26 @@ PYBIND11_MODULE(_simulation_abm, m)
...
+        .def_readwrite("type", &mio::abm::TestParameters::type);
+
+    pymio::bind_CustomIndexArray<mio::UncertainValue<double>, mio::abm::VirusVariant, mio::AgeGroup>(
+        m, "_AgeParameterArray");
+    pymio::bind_CustomIndexArray<mio::abm::TestParameters, mio::abm::TestType>(m, "_TestData");
+    pymio::bind_Index<mio::abm::ExposureType>(m, "ExposureTypeIndex");
+    pymio::bind_ParameterSet<mio::abm::ParametersBase, pymio::EnablePickling::Never>(m, "ParametersBase");
+    pymio::bind_class<mio::abm::Parameters, pymio::EnablePickling::Never, mio::abm::ParametersBase>(m, "Parameters")
+        .def(py::init<int>())
+        .def("check_constraints", &mio::abm::Parameters::check_constraints);
```
reneSchm, line 135: I wonder why this is marked private/hidden (the leading underscore)? If it is not meant to be accessed by the user, why do we bind it?

reneSchm (commented):
Why do we use ExposureType and TimePoint as pair for e.g. latest_exposure in the Infection ctor or the tests? I think using latest_exposure_time/_type as separate values is easier than using .first/,second everywhere, right?

Staying with the changes made in this branch, they look good to me, though there are merge conflicts with the newly merged #1072 (sorry) that have to be resolved before I can approve this.

pycode/memilio-simulation/memilio/simulation/bindings/models/abm.cpp
reneSchm: So the types are needed, e.g. when accessing the TestData Parameter. While the bindings build without issue when commenting out the _TestData bind, the test fails with the following message:

```
ERROR: test_locations (memilio.simulation_test.test_abm.TestAbm)
----------------------------------------------------------------------
TypeError: Unregistered type : mio::CustomIndexArray<mio::abm::TestParameters, mio::abm::TestType>

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/schm_r6/Documents/memilio/pycode/memilio-simulation/memilio/simulation_test/test_abm.py", line 60, in test_locations
    1), t0, t0 + abm.days(1), model.parameters.TestData[abm.TestType.Antigen], 1.0)
TypeError: Unable to convert function return value to a Python type! The signature was
        (arg0: memilio.simulation._simulation_abm.ParametersBase) -> mio::CustomIndexArray<mio::abm::TestParameters, mio::abm::TestType>
```

DavidKerkmann:
> Why do we use ExposureType and TimePoint as pair for e.g. latest_exposure in the Infection ctor or the tests? I think using latest_exposure_time/_type as separate values is easier than using .first/,second everywhere, right?

I think this got introduced when the function `std::pair<ExposureType, TimePoint> Person::get_latest_protection() const` got introduced, where the coupling was made and hence has been used like this everywhere. I agree separate value would be easier, but then we have to come up with something for this function, too. Any ideas?

reneSchm:
> I think this got introduced when the function `std::pair<ExposureType, TimePoint> Person::get_latest_protection() const` got introduced, where the coupling was made and hence has been used like this everywhere. I agree separate value would be easier, but then we have to come up with something for this function, too. Any ideas?

That makes sense. In that case, I think we could either split up the variables immediately after the function call, or we can add (yet another) simple struct for the protection. That way, we probably better preserve that the ExposureType is not used as (a person being) "exposed by", but rather as "protected from exposure by".

reneSchm (approved):
After the merge and additional renaming of Vaccine/ExposureType this is a very nice improvement.

PR #1106, opened 2024-08-12
1105 redesign bindings structure to improve typing

### LINKED ISSUE #1105: redesign bindings structure to improve typing

Currently our typing cannot resolve types defined in Simulation and then used in a model. This is due to our modules/submodules being built independently.

### Enhancement description

Restructure the way our bindings are compiled

### Additional context

_No response_

### Checklist

- [X] Linked to project

MaxBetzDLR:
# Changes and Information

- test different structures for building the bindings
- move bindings into subfolder
- fix bugs in bindings
- redesign stub generation, so more type hints work

### Checks by code author

Closes #1105


pycode/memilio-simulation/memilio/simulation/abm.py
```diff
@@ -22,4 +22,4 @@
 Python bindings for MEmilio ABM model.
 """
 
-from memilio._simulation_abm import *
+from memilio.simulation._simulation_abm import *
```
dabele, line 25: isn't there a conflict or duplication between the imports in abm.py, oseir.py, etc. and the ones in __get_attr__ in __init__.py? seems to me they do the same thing (I guess the __get_attr__ version has priority?)

pycode/memilio-simulation/memilio/simulation/bindings/simulation.cpp
```diff
@@ -141,7 +124,7 @@ PYBIND11_MODULE(_simulation, m)
             return std::vector<std::pair<mio::Date, mio::Date>>(h.begin(), h.end());
         },
         py::arg("state_id"), py::arg("start_date") = mio::Date(std::numeric_limits<int>::min(), 1, 1),
-        py::arg("end_date") = mio::Date(std::numeric_limits<int>::max(), 1, 1));
+        py::arg("end_date") = mio::Date(std::numeric_limits<int>::max(), 1, 1)); // TODO: Python doesnt know theses default values
```
dabele, line 127: can you elaborate what exactly doesn't work? does this need a new issue?

pycode/memilio-simulation/setup.py
```diff
@@ -25,14 +25,14 @@
...
-        # some excel files (e.g. population or twitter data)
-        'pandas>=2.0.0',
     ],
     extras_require={
         'dev': [
             # smaller numpy versions cause a security issue, 1.25 breaks testing with pyfakefs
             'numpy>=1.22,<1.25',
+            # smaller pandas versions contain a bug that sometimes prevents reading
+            # some excel files (e.g. population or twitter data)
+            'pandas>=2.0.0',
```
dabele, line 35: is pandas actually required by the simulation package? looks like this was just never amended when the epidata and simulation packages where split.

pycode/memilio-simulation/tools/generate_stubs.py
```diff
@@ -34,39 +37,113 @@
...
-        [python_interpreter, '-m', 'pybind11_stubgen', '--ignore-all-errors', '-o', output_dir, 'memilio._simulation'])
-    os.rename(os.path.join(output_module_dir, '_simulation.pyi'),
-              os.path.join(output_dir, '__init__.pyi'))
+        ['stubgen', '--include-docstrings', '-o', package_dir, '-p', 'memilio.simulation'])
+
+    # TODO: several error in the stubs (output of running mypy on codebase with stubs generated by mypy, need to set flag_configure_generated_stubs=False)
+    #   - fix numpy.float64[m, 1] to numpy.float64[M, Literal(1)] with M = typing.TypeVar("M", bound=int) at top of file
+    #   - or numpy.ndarray[numpy.float64[m, 1]] to numpy.ndarray[tuple[M, typing.Literal[1]], numpy.dtype[numpy.float64]]
+    #   - also numpy.ndarray[numpy.float64[m, n], flags.f_contiguous] needs to get n defined
+    #   - change following error:
```
dabele, line 61: I don't understand what this comment is supposed to tell me, at least the part in parantheses.

Also, is that something that requires a new issue?

pycode/memilio-simulation/CMakeLists.txt
```diff
@@ -32,6 +32,9 @@ endif()
 # add in C++ library
 add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/../../cpp ${CMAKE_CURRENT_BINARY_DIR}/cpp EXCLUDE_FROM_ALL)
 
+# add path to bindings
+include_directories(memilio/simulation/bindings)
```
dabele, line 36: why is that better than the `target_include_directories` you removed?

dabele (changes requested):

pycode/memilio-simulation/CMakeLists.txt
MaxBetzDLR: All of the individual targets use this include directories, so it seemed to make sense to set it for all targets once instead of calling the same function multiple times

pycode/memilio-simulation/setup.py
MaxBetzDLR: Pandas is used in the test and examples of simulation. Beccause of that pandas is part of the .[dev] target and not the base install requirements.

pycode/memilio-simulation/memilio/simulation/bindings/simulation.cpp
MaxBetzDLR: The old stub generator could not interpret the value of the instance of mio::Date(...). 
But with moving to mypy as the stub generator, we get no default values at all in the stubs of the c++ binded code. I didn't find a direct solution, that didn't include adjusting the stubs by hand after being generated.

pycode/memilio-simulation/memilio/simulation/abm.py
MaxBetzDLR: Both have a different use case and thus are not conflicting. When importing python only looks at the file structure and searches the exact file. **get_attr** is for the following case:
`import memilio.simulation as mio
mio.abm.Model()`
In this example **abm** is attribute of memilio.simulation and not searched by looking at the file system. To make that usable we need the **get_attr** function. But it should return the python modules and not the protected modules, because of possible pure python extensions

pycode/memilio-simulation/tools/generate_stubs.py
MaxBetzDLR: Improved the comment to be more comprehensive and moved parts of it to a different place. Those **TODOs* are low priority and need enough time to implement. I would move them into a future issue.

pycode/memilio-simulation/setup.py
dabele: I stumbled over that before, maybe it's not ideal to have that large dependency just for tests, but certainly not to be solved here.

pycode/memilio-simulation/memilio/simulation/abm.py
dabele: I see. If I understand correctly, your example isn't supported by normal python and has nothing to do with the c++ bindings themselves, right? 

Is it common in python projects to add __get_attr__ to make this work? And if so, shouldn't we add __get_attr__ at the level of memilio as well? It might be confusing if this works:
```
import memilio.simulation
memilio.simulation.abm.Model()
```
but this doesn't:
```
import memilio
memilio.simulation.abm.Model()
```

pycode/memilio-simulation/memilio/simulation/bindings/simulation.cpp
dabele: So is this related to the TODO in generate_stubs.py? 

as a general rule, every TODO comment in the code should have a corresponding issue in github to increase the chance of the TODO actually being done. Maybe one issue for all the TODOs is enough here if they are related?

pycode/memilio-simulation/CMakeLists.txt
dabele: `target_...` functions avoid global declarations that might contaminate other projects (same reason to avoid global variables and declarations in the global namespace). We had problems with this before. I guess it's unlikely this CmakeLists is included somewhere else, so it's a local problem. But in general use the `target_...` versions of cmake functions, unless you are absolutely certain that everyone using the file or adding to it requires the global setting and the normal `target_link_libraries` inheritance doesn't work for some reason.

pycode/memilio-simulation/memilio/simulation/__init__.py
```diff
@@ -27,19 +27,19 @@
 
 def __getattr__(attr):
```
dabele, line 28: please add a short comment here to explain the purpose.

dabele (changes requested):

pycode/memilio-simulation/memilio/simulation/abm.py
MaxBetzDLR: It depends on the project, eg. NumPy follows this convention. This is a decision we have to make. Implementing it at the level of memilio seems harder, because all of our submodules (simulation, epidata,..) have different __init__.py files (that do the same thing) and only one of those files is installed into the environment. 

Yes, this behavior could be confusing. On the other hand, memilio is more a collection of different packages inside a single namespaces, that work mostly indipendently. The submodules of simulation are more likely to be used together. 

Implementing it on level memilio does not seem possible, but i still find it usefull to have it for simulation

dabele: makes sense. We'll leave it as it is. 

It probably would work on memilio level, but each package would have to be listed in the `__init.py__` of each package, which is a hassle to maintain.

dabele (approved):

PR #1155, opened 2024-12-06
1154 integrator is handled as adaptive if using small fixed step sizes

### LINKED ISSUE #1154: Integrator is handled as adaptive if using small fixed step sizes

When calculating with a small but fixed step size, it can happen that the variable m_is_adaptive is set to true in the integrator, although no adaptive method is used. (I used a step size of 10^-2 and a high tmax).

### Enhancement description

It is sufficient to insert a tolerance in the calculation of m_is_adaptive.

### Additional context

_No response_

### Checklist

- [X] Linked to project

lenaploetzke:
# Changes and Information

- Added tolerance in the calculation of m_is_adaptive.
-
-

-
-

### Checks by code author

Closes #1154

### DIFF
--- a/cpp/memilio/config.h
+++ b/cpp/memilio/config.h
@@ -27,6 +27,7 @@
 
 #include "memilio/config_internal.h"
 #include <type_traits>
+#include "ad/ad.hpp"
 
 using ScalarType = double;
 
@@ -65,6 +66,15 @@ struct Limits<double> {
     static constexpr double zero_tolerance()
     {
         return 1e-12;
+    }
+};
+
+template <class AD_TAPE_REAL, class DATA_HANDLER>
+struct Limits<ad::internal::active_type<AD_TAPE_REAL, DATA_HANDLER>> {
+    /// @brief Returns the limit under which an ad::internal::active_type may be rounded down to zero.
+    static constexpr float zero_tolerance()
+    {
+        return Limits<AD_TAPE_REAL>::zero_tolerance();
     }
 };
 /** @} */
--- a/cpp/memilio/config.h
+++ b/cpp/memilio/config.h
@@ -35,7 +35,8 @@ namespace ad
 namespace internal
 {
 // Forward declaration used for support of ad types in Limits.
-template <class AD_TAPE_REAL, class DATA_HANDLER>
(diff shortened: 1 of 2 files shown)


cpp/memilio/config.h
```diff
@@ -27,6 +27,7 @@
 
 #include "memilio/config_internal.h"
 #include <type_traits>
+#include "ad/ad.hpp"
 
 using ScalarType = double;
```
reneSchm, line 33: Since we don't use any properties of the ad type we should get away with only a forward declaration. This way we avoid including the ad header in basically all files (because config.h is), which may reduce compilation time or binary size.

```suggestion

using ScalarType = double;

namespace ad
{
namespace internal
{
// Forward declaration used for support of ad types in Limits.
template <class AD_TAPE_REAL, class DATA_HANDLER>
struct active_type;
} // namespace internal
} // namespace ad
```

cpp/memilio/config.h
```diff
@@ -67,6 +68,15 @@ struct Limits<double> {
         return 1e-12;
     }
 };
+
+template <class AD_TAPE_REAL, class DATA_HANDLER>
+struct Limits<ad::internal::active_type<AD_TAPE_REAL, DATA_HANDLER>> {
+    /// @brief Returns the limit under which an ad::internal::active_type may be rounded down to zero.
```
reneSchm, line 74: I think using "AD values" or "AD value_type" instead of the internal name is more helpful in the documentation.

I am also a bit torn on the template names. As is, it is easy to find their original use and/or documentation, but they technically do not conform to our coding guidelines. I slightly lean to the conforming option, maybe use "AdReal" (or "AdTapeReal") and "DataHandler"?

reneSchm (changes requested):

cpp/memilio/config.h
lenaploetzke: That is a very good point, thank you!

cpp/memilio/config.h
lenaploetzke: Yes ok, I will adapt this to be in accordance with our guidelines.

lenaploetzke:
I don't know what the problem is with codecov again.

PR #1162, opened 2024-12-12
1159 add diffusive abm and smm

### LINKED ISSUE #1159: Add diffusive ABM and SMM
### Feature description

Add the diffusive abm including the quadwell potential and the smm from https://github.com/reneSchm/memilio/tree/example-hybrid-omp to main.

### Additional context

_No response_

### Checklist

- [X] Linked to project.

jubicker:
# Changes and Information

- add diffusive ABM (model, simulation, parameters and quadwell potential)
- add SMM (model, simulation, parameters)
- add tests for both models
- add examples for both models

### Checks by code author

closes #1159


cpp/models/d_abm/quad_well.h
```diff
@@ -0,0 +1,201 @@
...
+        , m_non_moving_states(non_moving_states)
+        , m_number_transitions(static_cast<size_t>(Status::Count), Eigen::MatrixXd::Zero(4, 4))
+    {
+        for (auto& agent : populations) {
+            mio::unused(agent);
+            assert(is_in_domain(agent.position));
+        }
+        for (auto& r : rates) {
+            m_adoption_rates.emplace(std::forward_as_tuple(r.region, r.from, r.to), r);
+            //m_adoption_rates[{r.region, r.from, r.to}] = r;
```
reneSchm, line 66:
```suggestion
```

cpp/models/d_abm/model.h
```diff
@@ -0,0 +1,52 @@
...
+    using Implementation::move;
+    using Implementation::time_point;
+
+    using Status = typename Implementation::Status;
+    using Agent  = typename Implementation::Agent;
+
+    inline constexpr void check_constraints() const
+    {
+    }
+};
```
reneSchm, line 47: Whoever would write all of this without a singular comment?
```suggestion

/**
 * @brief Wrap an implementation of a diffusive ABM so it can be run by the d_abm::Simulation.
 * Uses the CRTP. See comments on using statements for expected function signatures.
 * @tparam Implementation A class implementing all functions and types marked with the using keyword in Model.
 */
template <class Implementation>
class Model : public Implementation
{
public:
    /// Use the constructors defined by the Implementation
    using Implementation::Implementation;

    /**
     * @brief Set the status of an agent.
     * Expected signature: `void adopt(Agent&, const Status&)`
     */
    using Implementation::adopt;

    /**
     * @brief Calculate the current adoption rate of an agent from its status to the given one.
     * Expected signature: `double adoption_rate(const Agent&, const Status&)`
     */
    using Implementation::adoption_rate;

    /**
     * @brief Change the Position of an Agent, depending on its state, the current time and step size.
     * Expected signature: `void move(const double, const double, Agent&)`
     * The first argument is time, the second step size.
     */
    using Implementation::move;

    /**
     * @brief Get the Implementations RNG.
     * Expected signature: `mio::RandomNumberGenerator& get_rng()`
     */
    using Implementation::get_rng;

    /**
     * @brief Aggregate the population by their Status for the simulation result.
     * Expected signature: `Eigen::VectorXd time_point()`
     */
    using Implementation::time_point;

    /// @brief The status of an agent.
    using Status = typename Implementation::Status;
    /// @brief An agent is expected to contain at least a status and a position.
    using Agent = typename Implementation::Agent;

    /// @brief Empty function for compatability with MEmilio.
    inline constexpr void check_constraints() const
    {
    }
};```

**Review by reneSchm — CHANGES_REQUESTED** [review]
I seem to barely document code I write under a deadline. I will continue this with more suggestions later, but feel free to edit them - or add your own documentation, there is a lot to do...

The directory structure and naming looks pretty good. I have not checked the tests yet

File: cpp/tests/test_d_abm_model.cpp [hunk 3]
```diff
@@ -0,0 +1,142 @@
...
+    R,
+    D,
+    Count
+
+};
+
+TEST(TestQuadWell, well)
+{
+    auto well = well_index(Eigen::Vector2d{-1, 1});
+    EXPECT_EQ(well, size_t(0));
```
reneSchm, line 46: Test 1-3 as well?

cpp/tests/test_d_abm_model.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+    QuadWellModel<InfectionState>::Agent a1{Eigen::Vector2d{-1, 1}, InfectionState::S};
+    QuadWellModel<InfectionState>::Agent a2{Eigen::Vector2d{-1.2, 1}, InfectionState::I};
+    QuadWellModel<InfectionState> qw({a1, a2}, {{InfectionState::S,
+                                                 InfectionState::E,
+                                                 mio::dabm::Region(0),
+                                                 0.1,
+                                                 {InfectionState::C, InfectionState::I},
+                                                 {1, 0.5}}});
+    EXPECT_EQ(qw.adoption_rate(a1, InfectionState::E), 0.025);
```
reneSchm, line 68: Other branches that are not checked are
- the rate for a2 to E is 0
- another isolated infected does not influence the rate
- the rate of a1 scales correctly with more nearby infected and non-infected

cpp/tests/test_d_abm_model.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+    QuadWellModel<InfectionState>::Agent a1{Eigen::Vector2d{-1.2, 1}, InfectionState::S};
+    QuadWellModel<InfectionState>::Agent a2{Eigen::Vector2d{-1.2, 1}, InfectionState::I};
+    QuadWellModel<InfectionState> qw({a1, a2}, {}, 0.1, 0., {InfectionState::I});
+    qw.move(0, 0.1, a1);
+    qw.move(0, 0.1, a2);
+    EXPECT_EQ(a1.position[0], -0.9888);
+    EXPECT_EQ(a1.position[1], 1);
+    EXPECT_EQ(a2.position[0], -1.2);
+    EXPECT_EQ(a2.position[1], 1);
```
reneSchm, line 81: This should not work well without mocking, as move uses a normal distributed value.

cpp/models/d_abm/quad_well.h
```diff
@@ -0,0 +1,200 @@
...
+            std::find(m_non_moving_regions.begin(), m_non_moving_regions.end(), old_well) ==
+                m_non_moving_regions.end()) {
+            Position p = {mio::DistributionAdapter<std::normal_distribution<double>>::get_instance()(m_rng, 0.0, 1.0),
+                          mio::DistributionAdapter<std::normal_distribution<double>>::get_instance()(m_rng, 0.0, 1.0)};
+
+            agent.position      = agent.position - dt * grad_U(agent.position) + (m_sigma * std::sqrt(dt)) * p;
+            const auto new_well = well_index(agent.position);
+            if (old_well != new_well) {
+                m_number_transitions[static_cast<size_t>(agent.status)](old_well, new_well)++;
+            }
```
reneSchm, line 126: Do we want to keep these counters?

cpp/models/smm/parameters.h
```diff
@@ -0,0 +1,85 @@
...
+template <class Status>
+struct AdoptionRates {
+    using Type = std::vector<AdoptionRate<Status>>;
+    const static std::string name()
+    {
+        return "AdoptionRates";
+    }
+};
+
+template <class Status>
```
reneSchm, line 61:
```suggestion
/**
 * @brief Struct defining a possible regional transition in a Model based on Poisson Processes.
 * @tparam Status An infection state enum.
 */
template <class Status>
```

cpp/tests/test_d_abm_model.cpp
jubicker: It does because sigma is 0, therefore the movement is only defined by the function gradient.

reneSchm: Oh, I missed that. Can you add a comment that sigma=0? It is not that obvious from the constructor call.

cpp/models/d_abm/quad_well.h
jubicker: I would say yes. They might be helpful for output analysis.

cpp/tests/test_smm_model.cpp
```diff
@@ -156,16 +163,121 @@ TEST(TestSMMSimulation, advance)
...
+
+    //Simulate for 30 days
+    auto sim = mio::smm::Simulation(model, 0.0, 0.1);
+    sim.advance(30.);
+    //As model populations are all zero only t0 and tmax should be logged
+    EXPECT_EQ(sim.get_result().get_num_time_points(), 2);
     EXPECT_EQ(sim.get_result().get_last_time(), 30.);
 }
+
+TEST(TestSMMSimulation, covergence)
```
reneSchm, line 209:
```suggestion
TEST(TestSMMSimulation, convergence)
```

cpp/tests/test_smm_model.cpp
```diff
@@ -156,16 +163,121 @@ TEST(TestSMMSimulation, advance)
...
+    }
+    //The number of transitions from region 0 to region 1 should be approx. rate * pop in region 0
+    double rel_diff1 =
+        std::abs(rate * pop1 - std::abs(std::accumulate(transitions1.begin(), transitions1.end(), 0.0) / num_runs1)) /
+        (rate * pop1);
+    double rel_diff2 =
+        std::abs(rate * pop2 - std::abs(std::accumulate(transitions2.begin(), transitions2.end(), 0.0) / num_runs2)) /
+        (rate * pop2);
+
+    EXPECT_GE(rel_diff1, rel_diff2);
```
reneSchm, line 282: Why do you compare the two models with each other rather than with the expected result? The tolerance for comparing with the result is probably fairly large, but I think it would be better than comparing two "random" values.

Also, why the greater equal? If both models converge, reldiff1 < rel_diff2 should be possible, right? Maybe use one model, and compare the simulated rate against the expected rate, with some tolerance (with can be fairly large).

reneSchm: Moreover, should this be a random test perhaps? So that it checks not only that one specific seed works. It is okay for a random test to possibly fail, if it is very unlikely that it does. You can run it a couple 1000 times to check that the tolerance is large enough for most edge cases.

reneSchm (changes requested):

cpp/tests/test_d_abm_model.cpp
reneSchm: Can you add descriptions to the test what it is you check? Otherwise it is hard to tell what is going on here.

This is true for all added tests. For example, in TestQuadWell.well(_index ?) one comment at the top like "check that well_index returns indices in the correct order" would be enough.
Here, in TestQuadWell.adoptionRate, you could write down what the agents and rates are used for, and/or explain what the EXPECTs do. Like "check that a1 is affected by a2, but not the for off a3" for the first EXPECT call, and so on.

cpp/tests/test_smm_model.cpp
jubicker: What I do here is check whether the simulated transition rate converges to the expected transition rate for num_runs, pop_size to infty. And I realize this by checking whether the relative difference between the simulated number of transitions and the expected number of transitions gets smaller if I increase the number of runs as well as the number of agents.

I can also only use one model and compare the simulated and expected number of transitions, but what would be an appropriate tolerance? 1%, 50%? An how many runs do we require to meet that tolerance?

cpp/memilio/epidemiology/adoption_rate.h
```diff
@@ -0,0 +1,59 @@
...
+ * the sum over all "influences" scaled by their respective "factors".
+ * @tparam Status An infection state enum.
+ */
+template <class Status>
+struct AdoptionRate {
+    Status from; // i
+    Status to; // j
+    mio::regions::Region region; // k
+    ScalarType factor; // gammahat_{ij}^k
+    std::vector<Influence<Status>> influences;
```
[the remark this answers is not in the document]
jubicker: Now we do

cpp/models/d_abm/parameters.h
```diff
@@ -21,41 +21,10 @@
...
-};
-
-template <class Status>
-struct AdoptionRates {
-    using Type = std::vector<AdoptionRate<Status>>;
-    const static std::string name()
-    {
-        return "AdoptionRates";
-    }
-};
```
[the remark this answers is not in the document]
jubicker: We decided to keep this file and add a comment on why it is empty.

cpp/models/d_abm/quad_well.h
```diff
@@ -0,0 +1,270 @@
...
+                m_non_moving_states.end() &&
+            std::find(m_non_moving_regions.begin(), m_non_moving_regions.end(), old_well) ==
+                m_non_moving_regions.end()) {
+            Position p = {mio::DistributionAdapter<std::normal_distribution<double>>::get_instance()(m_rng, 0.0, 1.0),
+                          mio::DistributionAdapter<std::normal_distribution<double>>::get_instance()(m_rng, 0.0, 1.0)};
+
+            agent.position      = agent.position - dt * grad_U(agent.position) + (m_sigma * std::sqrt(dt)) * p;
+            const auto new_well = well_index(agent.position);
+            if (old_well != new_well) {
+                m_number_transitions[static_cast<size_t>(agent.status)](old_well, new_well)++;
```
reneSchm, line 159: Can you add an agent (or change ones position) in the move test, so this line gets covered? Maybe check as well that number transitions = 0,...,0,1,0,...,0

cpp/tests/test_smm_model.cpp
```diff
@@ -242,42 +259,14 @@ TEST(TestSMMSimulation, covergence)
...
-        (rate * pop1);
-    double rel_diff2 =
-        std::abs(rate * pop2 - std::abs(std::accumulate(transitions2.begin(), transitions2.end(), 0.0) / num_runs2)) /
-        (rate * pop2);
-
-    EXPECT_GE(rel_diff1, rel_diff2);
+    double mean_trans = std::accumulate(transitions.begin(), transitions.end(), 0.0) / num_runs;
+    double rel_dev    = std::abs(expected_num_trans - mean_trans) / expected_num_trans;
+    //The relative deviation should not exceed 15%
+    EXPECT_LE(rel_dev, 0.15);
```
reneSchm, line 271: Nice, 15% is honestly better than I expected.

reneSchm (approved):

PR #1168, opened 2024-12-19
1167 Update some dependencies

### LINKED ISSUE #1167: Update some dependencies

A recent CI failure is caused by gcc 14 having issues with the particular spdlog version we use. This can be fixed by "simply" updating to a newer version.

### Enhancement description

Update spdlog version, and optionally other dependencies as well.

### Additional context

_No response_

### Checklist

- [ ] Linked to project

reneSchm:
# Changes and Information

- Update minimum required cmake 3.11 -> 3.13
- Update spdlog 1.11.0 -> 1.15.0
- Update Eigen 3.3.9 -> 3.4.0
- Update JsonCpp 1.9.5 -> 1.9.6

- These updates help with support of recent compiler versions 
- cmake 3.13 was already required by install in cpp/CMakeLists.txt:183, prior versions are unable to install targets from a subdirectory 
- A newer spdlog version is required unless we want to silence several compiler warnings when including the library with gcc 14+. chose newest release, though we may use an older one if there is a reason for it
- Eigen 3.4.0 is newest release, but already 2 years old. It adds more compatibility with STL, e.g. iterators and initializer list support.
  - **IMPORTANT**: now including Eigen as system library to avoid several compiler warnings from within the library. could use several pragmas to silence each error individually instead
  - Eigen 4.0.0 may be available soon<sup>(TM)</sup>
- JsonCpp 1.9.6 has several bugfixes, though we are not affected by these bugs (AFAIK)
- Check commits for steps taken to update the code to these versions. Updates are independent and can be selectively undone, if desired 
  - changes to the serialization test are due to stricter is_container requirement, causing std::array to be serialized as tuple. deserialization of std::array was not supported anyways, so this is not a breaking change
 
### Checks by code author

Closes #1167

### DIFF
--- a/cpp/memilio/math/stepper_wrapper.h
+++ b/cpp/memilio/math/stepper_wrapper.h
@@ -24,6 +24,7 @@
 #include "memilio/utils/logging.h"
 
 #include "boost/numeric/odeint/external/eigen/eigen_algebra.hpp"
+#include "boost/numeric/odeint/external/eigen/eigen_resize.hpp"
 #include "boost/numeric/odeint/stepper/controlled_runge_kutta.hpp"
 #include "boost/numeric/odeint/stepper/runge_kutta_fehlberg78.hpp"
 #include "boost/numeric/odeint/stepper/runge_kutta_cash_karp54.hpp"
@@ -44,7 +45,7 @@ class ControlledStepperWrapper : public mio::IntegratorCore<FP>
     using Stepper = boost::numeric::odeint::controlled_runge_kutta<
         ControlledStepper<Vector<FP>, FP, Vector<FP>, FP, boost::numeric::odeint::vector_space_algebra,
                           typename boost::numeric::odeint::operations_dispatcher<Vector<FP>>::operations_type,
-                          boost::numeric::odeint::never_resizer>>;
+                          boost::numeric::odeint::initially_resizer>>;
     static constexpr bool is_fsal_stepper = std::is_same_v<typename Stepper::stepper_type::stepper_category,
                                                            boost::numeric::odeint::explicit_error_stepper_fsal_tag>;
     static_assert(!is_fsal_stepper,
@@ -109,7 +110,6 @@ public:
                 step_result = m_stepper.try_step(
                     // reorder arguments of the DerivFunction f for the stepper
                     [&](const Vector<FP>& x, Vector<FP>& dxds, FP s) {
-                        dxds.resizeLike(x); // boost resizers cannot resize Eigen::Vector, hence we need to do that here
                         f(x, s, dxds);
                     },
                     m_yt, t, m_ytp1, dt);
@@ -186,7 +186,7 @@ class ExplicitStepperWrapper : public mio::IntegratorCore<FP>
 public:
     using Stepper = ExplicitStepper<Vector<FP>, FP, Vector<FP>, FP, boost::numeric::odeint::vector_space_algebra,
                                     typename boost::numeric::odeint::operations_dispatcher<Vector<FP>>::operations_type,
-                                    boost::numeric::odeint::never_resizer>;
+                                    boost::numeric::odeint::initially_resizer>;
 
     /**
      * @brief Set up the integrator.
@@ -213,7 +213,6 @@ public:
         m_stepper.do_step(
             // reorder arguments of the DerivFunction f for the stepper
             [&](const Vector<FP>& x, Vector<FP>& dxds, FP s) {
-                dxds.resizeLike(x); // do_step calls sys with a vector of size 0 for some reason
                 f(x, s, dxds);
             },
             ytp1, t, dt);
--- a/cpp/memilio/math/stepper_wrapper.h
+++ b/cpp/memilio/math/stepper_wrapper.h
@@ -43,8 +43,8 @@ template <typename FP,
 class ControlledStepperWrapper : public mio::IntegratorCore<FP>
 {
     using Stepper = boost::numeric::odeint::controlled_runge_kutta<
-        ControlledStepper<Vector<FP>, FP, Vector<FP>, FP, boost::numeric::odeint::vector_space_algebra,
-                          typename boost::numeric::odeint::operations_dispatcher<Vector<FP>>::operations_type,
+        ControlledStepper<Eigen::VectorX<FP>, FP, Eigen::VectorX<FP>, FP, boost::numeric::odeint::vector_space_algebra,
+                          typename boost::numeric::odeint::operations_dispatcher<Eigen::VectorX<FP>>::operations_type,
                           boost::numeric::odeint::initially_resizer>>;
     static constexpr bool is_fsal_stepper = std::is_same_v<typename Stepper::stepper_type::stepper_category,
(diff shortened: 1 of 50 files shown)


mknaranja:
@reneSchm Please update readmes :)

cpp/memilio/utils/logging.h
```diff
@@ -127,38 +127,24 @@ inline void log(LogLevel level, spdlog::string_view_t fmt, const Args&... args)
...
-    constexpr auto parse(format_parse_context& ctx) -> decltype(ctx.begin())
-    {
-        return ctx.end();
-    }
-    template <typename FormatContext>
-    auto format(const ad::gt1s<double>::type& input, FormatContext& ctx) -> decltype(ctx.out())
-    {
-        return format_to(ctx.out(), "{}", ad::value(input));
-    }
-};
```
mknaranja, line 145: don't see why so much became superfluous?!

cpp/tests/abm_helpers.cpp
```diff
@@ -49,14 +49,16 @@ void interact_testing(mio::abm::PersonalRandomNumberGenerator& personal_rng, mio
 {
     // allocate and initialize air exposures with 0
     mio::abm::AirExposureRates local_air_exposure;
-    local_air_exposure.resize({mio::abm::CellIndex(location.get_cells().size()), mio::abm::VirusVariant::Count});
```
mknaranja, line 52: have you checked that there are no other relevant resizes that need to be changed?

cpp/memilio/utils/uncertain_value.h
```diff
@@ -227,6 +227,15 @@ class UncertainValue
     std::unique_ptr<ParameterDistribution> m_dist;
 };
 
+/**
+ * @brief Format UncertainValues using their value for logging with spdlog.
+ */
+template <class FP>
+double format_as(const UncertainValue<FP>& uv)
```
mknaranja, line 234: don't we then also need a format_as for an ad type with UV and derivative?

mknaranja (changes requested):
Thank you for the updates. I gave some more or less informed comments. Maybe check for them, update the readmes and then it should be OK.

cpp/memilio/math/stepper_wrapper.h
```diff
@@ -44,7 +45,7 @@ class ControlledStepperWrapper : public mio::IntegratorCore<FP>
     using Stepper = boost::numeric::odeint::controlled_runge_kutta<
         ControlledStepper<Vector<FP>, FP, Vector<FP>, FP, boost::numeric::odeint::vector_space_algebra,
                           typename boost::numeric::odeint::operations_dispatcher<Vector<FP>>::operations_type,
-                          boost::numeric::odeint::never_resizer>>;
+                          boost::numeric::odeint::initially_resizer>>;
```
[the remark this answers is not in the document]
reneSchm: This replaces the line `dxds.resizeLike(x);` in the step method. The initially_resizer used to fail (silently) to resize the vector with the older Eigen version. But after the update, the resizers (in part. initially_resize) defined in the newly included boost header (line 27, eigen_resize.hpp) work as intended.

There are never_resize, initially_resize and always_resize. When try_step is called, boost resizes x and dxds never, the first time, or always respectively. I assume here that the vector sizes should not change, as the result TimeSeries would have issues with that anyways, hence I chose initially_resize.

cpp/memilio/math/stepper_wrapper.h
```diff
@@ -109,7 +110,6 @@ class ControlledStepperWrapper : public mio::IntegratorCore<FP>
                 step_result = m_stepper.try_step(
                     // reorder arguments of the DerivFunction f for the stepper
                     [&](const Vector<FP>& x, Vector<FP>& dxds, FP s) {
-                        dxds.resizeLike(x); // boost resizers cannot resize Eigen::Vector, hence we need to do that here
```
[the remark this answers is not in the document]
reneSchm: I did not update boost, but the "new" Eigen version (it is 2 years old now) works better with boost, see response to the resizers above.

cpp/tests/abm_helpers.cpp
reneSchm: Yes, we don't resize CustomIndexArray that much anyways. Also, if it were used incorrectly somewhere, the compiler would complain, vehemently.

cpp/memilio/utils/custom_index_array.h
```diff
@@ -235,8 +235,9 @@ class CustomIndexArray
     }
 
     /**
-     * Resize all dimensions.
-     * @param new_dims new dimensions.
+     * @brief Resize all dimensions.
+     ' Note that when increasing the overall size, new values may be uninitialized.
```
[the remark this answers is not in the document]
reneSchm: The internal array is stored linearly and reduced (or extended) to the product of all dimensions. So your 3x3 matrix is really just an array (1 2 3 4 5 6 7 8 9), with matrix accessor. So resizing to 1x9 or 9x1 would leave the array untouched. If we cut that matrix to 
- 3x2, we get 
1 2
3 4
5 6 
- 3x1, we get 
1
2
3
- 2x3, we get
1 2 3
4 5 6
- 2x2, we get
1 2
3 4

cpp/memilio/utils/uncertain_value.h
reneSchm: Good catch, that "double" should be a "FP" (or better "const FP&"). fmt is able to call the AD format_as if an AD type is returned here.

cpp/memilio/utils/logging.h
reneSchm: Two things. First, since we only print the value part of the AD type, I use a simpler part of the [fmt API](https://fmt.dev/latest/api/#formatting-user-defined-types), i.e. format_as,  to do essentially the same as this struct in fewer lines.
Second, I used the same template deduction @lenaploetzke introduced in #1155 to handle both (probably all) AD types, hence there is only one function replacing both types.

reneSchm:
Thanks for the review @mknaranja. Please check the "additional information for reviewers", in case you haven't.

I had made Eigen a system library, so we do not need to manually add compiler diagnostics. I hadn't removed them before your review ([see here](https://github.com/SciCompMod/memilio/pull/1168/commits/46d0ef9ca9cff999280700adf02f5b1402344ffb#diff-f86e6cb1e4100691d5ccc19e29cb4d613d8a8dd92c4e336170edce5324041f0cR24)).

I also replaced mio::Vector by Eigen::VectorX (new in 3.4.0) in that commit, so it is more obvious that Vector is a Eigen type and not something we wrote ourselves.

cpp/memilio/utils/custom_index_array.h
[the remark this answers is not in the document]
mknaranja: I was expecting that. Is that really expected behavior for a (new) user? Shouldn't we avoid sizing down with keeping the values as this is normally not what you what want when reducing dimensions?

mknaranja (commented):
Looks good. There is just one comment / discussion we should agree on.

mknaranja (approved):

PR #1179, opened 2025-01-09
17 Timing framework

### LINKED ISSUE #17: Time measurement
In order to better detect bottlenecks or time consuming code parts, time measurement should be enabled / integrated. The question remains how we would like to do this. A simple option would be via OpenMP but this would be that it gets deactivated once the project is not build with OpenMP by a user.

Any better suggestions @dabele @joergbrech ?

(Text added by @mknaranja)

reneSchm:
# Changes and Information

- Update to C++20
- Add a framework with an thread safe Timer.
- Used via ScopedTimer, it should guarantee that the timers use the correct thread.
- Timers are aggregated via TimeRegistrar, and can be evaluated.

- What is missing from the framework:
  - A way to disable timers. Disabling during runtime is intentionally not implemented, as it creates some overhead for start/stop methods. A compile time disable could be implemented using e.g. the Tag.
  - CPU timers. The current impl. only supports wall clock (std::steady_clock or omp_get_wtime).
  - Instancing the same Timer multiple times. This is a deliberate trade-off, as using static thread_local Timers enables thread safety and using the CT map.     
- Please check the naming. In particular, an alternative for ScopedTimer would be AutoTimer, as the prefix "auto" is somewhat commonly used to describe classes that use their lifetime as switch on/off (e.g. auto_mutex). Also the TimerRegistrar could perhaps use a new name.

### Checks by code author

Closes #17


cpp/memilio/timer/definitions.cpp
```diff
@@ -0,0 +1,72 @@
...
+ * @return The duration in seconds.
+ */
+double time_in_seconds(DurationType duration)
+{
+    return std::chrono::duration_cast<std::chrono::duration<double>>(duration).count();
+}
+
+/**
+ * @brief Get the current time.
+ * @return Returns omp_get_wtime() if OpenMP is enabled, stead_clock::now() otherwise.
```
HenrZu, line 62:
```suggestion
 * @return Returns omp_get_wtime() if OpenMP is enabled, steady_clock::now() otherwise.
```

cpp/memilio/timer/definitions.h
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * @brief Convert a duration to a (floating point) number of seconds.
+ * @param[in] duration Any DurationType value, mainly `BasicTimer::get_elapsed_time()`.  
+ * @return The duration in seconds.
+ */
+double time_in_seconds(DurationType duration);
+
+/**
+ * @brief Get the current time.
+ * @return Returns omp_get_wtime() if OpenMP is enabled, stead_clock::now() otherwise.
```
HenrZu, line 54:
```suggestion
 * @return Returns omp_get_wtime() if OpenMP is enabled, steady_clock::now() otherwise.
```

cpp/memilio/timer/registration.h
```diff
@@ -0,0 +1,62 @@
...
+
+struct TimerRegistration {
+    std::string name;
+    std::string scope;
+    BasicTimer& timer;
+    int thread_id;
+};
+
+/**
+ * @brief If scope is empty, returns name. Otherwise, concatonates scope, two colons "::" and name.
```
HenrZu, line 41:
```suggestion
 * @brief If scope is empty, returns name. Otherwise, concatenates scope, two colons "::" and name.
```

cpp/memilio/timer/table_printer.h
```diff
@@ -0,0 +1,273 @@
...
+     * @param[inout] out The stream to write to.
+     */
+    void print(const std::list<TimerRegistration>& timer_register, std::ostream& out = std::cout) override
+    {
+        auto [table, is_multithreaded, name_width, max_val] = create_table(timer_register);
+        // calculate column_widths
+        const size_t time_width   = fmt::format(fmt::runtime(m_time_format), max_val).size();
+        const size_t thread_width = table.get_col_name(table.cols() - 1).size();
+        std::vector<size_t> col_widths(table.cols() + 1);
+        // note that col_width is offset by 1 relative to table, as uses index 0 for the first value,
```
HenrZu, line 142:
```suggestion
        // note that col_width is offset by 1 relative to table, as it uses index 0 for the first value,
```

cpp/memilio/timer/table_printer.h
```diff
@@ -0,0 +1,273 @@
...
+
+        draw_hline();
+        // table header, prints table and column names
+        out << border_l << std::setw(col_widths[0]) << std::left << table.get_name();
+        for (size_t col = 0; col < table.cols(); col++) {
+            out << separator << std::setw(col_widths[col + 1]) << std::left << table.get_col_name(col);
+        }
+        out << border_r;
+        draw_hline();
+        // print table content, only adding statistics when mulithreaded
```
HenrZu, line 171:
```suggestion
        // print table content, only adding statistics when multithreaded
```

cpp/memilio/timer/table_printer.h
```diff
@@ -0,0 +1,273 @@
...
+                out << separator << std::setw(thread_width) << std::right << num_threads;
+            }
+            out << border_r;
+        }
+        draw_hline();
+    }
+
+private:
+    /**
+     * @brief The first part of the print function, separated as a somewhat independant step.
```
HenrZu, line 195:
```suggestion
     * @brief The first part of the print function, separated as a somewhat independent step.
```

cpp/memilio/timer/timer_registrar.h
```diff
@@ -0,0 +1,163 @@
...
+            {
+                // print_timers already uses a "single nowait", but the cout for titling the table needs one, too
+                std::cout << "Final Timer Summary\n";
+                print_timers();
+            }
+        }
+    }
+
+    std::unique_ptr<Printer> m_printer = std::make_unique<TablePrinter>(); ///< A printer to visualize all timers.
+    bool m_print_on_death              = true; ///< Whether to call m_printer during the desctructor.
```
HenrZu, line 155:
```suggestion
    bool m_print_on_death              = true; ///< Whether to call m_printer during the destructor.
```

HenrZu (changes requested):
The code is incredibly good and well structured. I only have one comment in the BasicTimer. Otherwise, i have read through the whole doc and found a few typos. The doc helped a lot to understand the structure. Great work :)


A readme in the time dir would be helpful. Tests and Readthedocs documentation is also missing (as you said before :) )

cpp/memilio/timer/basic_timer.h
```diff
@@ -0,0 +1,67 @@
...
+{
+namespace timing
+{
+
+/// @brief A minimal timer class.
+class BasicTimer
+{
+public:
+    /// @brief Start the timer. Must be followed by exactly one stop.
+    void start()
```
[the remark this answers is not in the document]
reneSchm: I avoided logic like `if (is_running)` or `if (is_enabled)` inside the start/stop functions on purpose, to keep it fast. But I could add a variable and checks for debug builds only, and put asserts or log_criticals when used incorrectly. Is that sufficient?

cpp/memilio/timer/auto_timer.h
```diff
@@ -0,0 +1,74 @@
...
+public:
+    /// @brief Run the NamedTimer given by the template parameter(s) Name (and Scope).
+    AutoTimer()
+        : m_timer(NamedTimer<Name, Scope>::get_instance())
+    {
+        m_timer.start();
+    }
+
+    /// @brief Run the given BasicTimer. Does not take ownership, so mind the timer's lifetime!
+    AutoTimer(BasicTimer& timer)
```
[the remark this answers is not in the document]
reneSchm: That would "fix" the lifetime issue - but I don't really want to do that. AutTimer's job is to call start and stop, NamedTimer takes care of storage and threading, BasicTimer defines the basic functionality. I would rather not mix these responsibilities, especially since the shared_ptr would not work well with how get_instance works.

Also, since you have to pass a lvalue reference, it should be fairly hard to invalidate the reference during timing. Something like `AutoTimer t(BasicTimer{})` does not work here.

cpp/memilio/timer/definitions.cpp
reneSchm: Well spotted. I will remove documentation from the .cpp, and only fix this in the .h file.

cpp/memilio/timer/basic_timer.h
```diff
@@ -34,31 +36,52 @@ class BasicTimer
...
+    {
+        should_be_running(false, "~BasicTimer");
+    }
+
 private:
     TimeType m_start_time; ///< The last time point at which start() was called
     DurationType m_elapsed_time{0}; ///< The total time spent between starts and stops.
+
+#ifndef NDEBUG
+    bool m_is_running = false; ///< In Debug builds, tracks whether the timer is running.
```
HenrZu, line 77: a public getter might be nice in debug mode

docs/source/cpp/performance_timers.rst
```diff
@@ -0,0 +1,204 @@
...
+
+        {
+            mio::timer::AutoTimer<"my simulation"> my_timer; // my_timer starts here
+            sim.advance(t_max); // run the simulation
+        } // my_timer stops here
+
+        ... // evaluate results
+    }
+
+and will see a table printed at the end of your programm, that lists the time it took to :code:`advance` next to the
```
HenrZu, line 42:
```suggestion
and will see a table printed at the end of your program, that lists the time it took to :code:`advance` next to the
```

docs/source/cpp/performance_timers.rst
```diff
@@ -0,0 +1,204 @@
...
+You can add more timers like this, but make sure you use unique names, otherwise the same timer will be reused, and the
+measured times will be added together. The name of the timer object itself (here :code:`my_timer`) is not important, as
+long as the compiler does not complain about it.
+
+
+Timing in the library
+~~~~~~~~~~~~~~~~~~~~~
+
+Adding timers in the library is not much different to adding timers in main, but avoiding name collisions can be more
+difficult. Hence, we use the optional second tempalte argument of AutoTimer to specify its scope, as shown in the
```
HenrZu, line 54:
```suggestion
difficult. Hence, we use the optional second template argument of AutoTimer to specify its scope, as shown in the
```

docs/source/cpp/performance_timers.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Or, when timing a free function:
+
+.. code-block:: cpp
+
+    #include "memilio/timer/auto_timer.h"
+
+    namespace foo {
+
+    void bar {
```
HenrZu, line 83:
```suggestion
    void bar() {
```

docs/source/cpp/performance_timers.rst
```diff
@@ -0,0 +1,204 @@
...
+Classes and their responsibilities
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In this section, we describe the main components of the timing framework and how they interact. For more details on a
+specific component, view its API documentation. 
+
+- **BasicTimer**:
+  The foundation of the timing framework. BasicTimer is a very simple class, that defines the methods start, stop,
+  reset, and get_elapsed_time. These are used by all other classes in this framework. Uses a wall clock, so if compute
+  resources are shared with other tasks, the timing results may be higher than expexted. In debug builds, it will log
```
HenrZu, line 141:
```suggestion
  resources are shared with other tasks, the timing results may be higher than expected. In debug builds, it will log
```

HenrZu (changes requested):
Looks great :) Only short comments/suggestions. 

I'm still unsure about the namespace, which is better :D But we should keep it consistent. So either we name the directory in timing, or the namespace in mio::time.

HenrZu:
i just linked the issue for the bump of the c++ version. Could you also update the c++ version here:

- docs/source/development.rst
- pycode/memilio-generation/memilio/generation/ast.py

cpp/memilio/timer/basic_timer.h
reneSchm: It is usually a bad idea to have the interface depend on compiler flags, so either we add the getter and update is_running in Release as well, or we keep it inaccessible (other than through log messages).

reneSchm: I would prefer keeping it inaccessible, since updating it without checks seems odd, and checking it is more expensive than I'd like.
Also, why would you need to check that the timer is running? I can only think of bad use cases. 
We can always expose the variable later, when we find a good use for it.

PR #1211, opened 2025-03-07
978 Local Feedback Mechanism 

HenrZu:
# Changes and Information

- Add a local Feedback Mechanism based on Dönges et al. (doi.org/10.3389/fphy.2022.842180)
- The Extension of this approach for metapopulations (see doi.org/10.48550/arXiv.2502.14428) will be adated, once this here is merged

-
-

### Checks by code author


cpp/memilio/compartments/feedback_simulation.h
```diff
@@ -0,0 +1,370 @@
...
+    {
+        return "NominalICUCapacity";
+    }
+};
+
+template <typename FP = double>
+using ParametersFeedback = ParameterSet<ICUOccupancyLocal<FP>, GammaShapeParameter<FP>, GammaScaleParameter<FP>,
+                                        GammaCutOff, ContactReductionMax<FP>, ContactReductionMin<FP>,
+                                        SoftPlusCurvatureParameter<FP>, NominalICUCapacity<FP>>;
+
```
reneSchm, line 163: I think "FeedbackParameters" or "FeedbackSimulationParameters" would sound nicer, but I don't mind this name too much.

Should we maybe have a `feedback_parameters.h`? Since most other ParameterSets use their own header.

cpp/memilio/compartments/feedback_simulation.h
```diff
@@ -0,0 +1,370 @@
...
+{
+public:
+    /**
+     * @brief Constructs the FeedbackSimulation by using a existing simulation instance and ICU compartment indices.
+     *
+     * @param sim The simulation instance to be extended with feedback mechanism.
+     * @param icu_indices A vector of indices indicating ICU compartments for specific model.
+     */
+    explicit FeedbackSimulation(Sim sim, const std::vector<size_t>& icu_indices)
+        : m_simulation(std::move(sim))
```
reneSchm, line 186: Using both move and pass-by-value is a bit odd, and looking at the example, it seems like you want FbSim to take ownership of sim. In that case, the constructor should take a rvalue ref, i.e. `Sim&& sim`. This also should make the std::move in line 186 here superfluous.

cpp/memilio/compartments/feedback_simulation.h
```diff
@@ -0,0 +1,370 @@
...
+        , m_feedback_parameters(m_simulation.get_model().parameters.get_num_groups())
+        , m_perceived_risk(static_cast<size_t>(m_simulation.get_model().parameters.get_num_groups()))
+    {
+    }
+
+    /**
+     * @brief Advances the simulation until tmax while applying feedback at fixed intervals.
+     *
+     * The simulation is advanced in steps of dt_feedback. At each step, feedback
+     * is applied, then the simulation is advanced, and afterwards the current ICU occupancy is stored.
```
reneSchm, line 197: Please add that Sim may make additional substeps depending on its dt. For fixed steppers, dt_feedback should be an integer multiple of dt.

reneSchm (changes requested):
Took a quick look at the overall structure, but not the calculations or test contents. Seems pretty good otherwise.

cpp/examples/ode_secir_feedback.cpp
```diff
@@ -0,0 +1,123 @@
...
+    // set up the parameters for the feedback simulation
+    initialize_feedback(feedback_simulation);
+
+    // run the simulation with feedback mechanism
+    auto results = feedback_simulation.advance(tmax);
+
+    // print the perceived risk and the final total population
+    auto& perceived_risk = feedback_simulation.get_perceived_risk();
+    perceived_risk.print_table({"Perceived Risk"});
+    std::cout << "Population at tmax = " << results.sum() << std::endl;
```
jubicker, line 121: Why do you print this as a result? It should just be the total_population you initialized, right?

cpp/memilio/compartments/feedback_simulation.h
```diff
@@ -0,0 +1,375 @@
...
+#include "memilio/utils/uncertain_value.h"
+#include "memilio/epidemiology/damping_sampling.h"
+
+namespace mio
+{
+/**
+ * @brief Daily local ICU occupancy aggregated for each age group.
+ */
+template <typename FP>
+struct ICUOccupancyLocal {
```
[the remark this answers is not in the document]
HenrZu: Originally, I wanted to put this PR directly together the spatial resolved feedback. But that would have been too much. That's where the naming came from. I have adapted it to `ICUOccupancyHistory`

cpp/examples/ode_secir_feedback.cpp
HenrZu: the `result` was the return from the call `feedback_simulation.advance(tmax);`. My idea was that this would be a good point for interested people to get access to the results, while not showing the results apart from the perceived risk. However, this way the result was unused and would cause an error in the creation process. Therefore, I printed the total population :)
Now I have removed the result variable.

jubicker (approved):

reneSchm (approved):

PR #1235, opened 2025-03-24
1178 code creation based on standardised formats for ODEs

### LINKED ISSUE #1178: Code creation based on standardised formats for ODEs
### Feature description

We want to automatically create the code for simulating an ODE model given in a standardised format like [Kappa](https://kappalanguage.org/) or [SBML](https://sbml.org/). 

### Additional context

_No response_

### Checklist

- [X] Linked to project.

kilianvolmer:
# Changes and Information

- Add memilio2sbml executable
- Add documentation for memilio2sbml executable

### Checks by code author

Closes #1178

### DIFF
--- a/cpp/memilio/CMakeLists.txt
+++ b/cpp/memilio/CMakeLists.txt
@@ -40,6 +40,8 @@ add_library(memilio
     io/history.h
     io/mobility_io.h
     io/mobility_io.cpp
+    io/parameters_io.h
+    io/parameters_io.cpp
     io/result_io.h
     io/result_io.cpp
     io/epi_data.h
--- a/cpp/memilio/io/parameters_io.cpp
+++ b/cpp/memilio/io/parameters_io.cpp
@@ -1,0 +1,72 @@
+/* 
+* Copyright (C) 2020-2025 MEmilio
+*
+* Authors: Henrik Zunker
+*
+* Contact: Martin J. Kuehn <Martin.Kuehn@DLR.de>
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#include "memilio/config.h"
+
+#ifdef MEMILIO_HAS_JSONCPP
+
+#include "memilio/io/epi_data.h"
+#include "memilio/io/result_io.h"
+#include "json/value.h"
+#include <string>
+#include <vector>
+
+namespace mio
+{
+IOResult<std::vector<std::vector<double>>> read_population_data(const std::vector<PopulationDataEntry>& population_data,
+                                                                const std::vector<int>& vregion)
+{
+    std::vector<std::vector<double>> vnum_population(
+        vregion.size(), std::vector<double>(ConfirmedCasesDataEntry::age_group_names.size(), 0.0));
+
+    for (auto&& county_entry : population_data) {
+        //accumulate population of states or country from population of counties
+        if (!county_entry.county_id && !county_entry.district_id) {
+            return failure(StatusCode::InvalidFileFormat, "File with county population expected.");
+        }
+        //find region that this county belongs to
+        //all counties belong to the country (id = 0)
+        auto it = std::find_if(vregion.begin(), vregion.end(), [&county_entry](auto r) {
+            return r == 0 ||
+                   (county_entry.county_id &&
+                    regions::StateId(r) == regions::get_state_id(int(*county_entry.county_id))) ||
+                   (county_entry.county_id && regions::CountyId(r) == *county_entry.county_id) ||
+                   (county_entry.district_id && regions::DistrictId(r) == *county_entry.district_id);
+        });
+        if (it != vregion.end()) {
+            auto region_idx      = size_t(it - vregion.begin());
+            auto& num_population = vnum_population[region_idx];
+            for (size_t age = 0; age < num_population.size(); age++) {
+                num_population[age] += county_entry.population[AgeGroup(age)];
+            }
+        }
+    }
+
+    return success(vnum_population);
+}
+
+IOResult<std::vector<std::vector<double>>> read_population_data(const std::string& path,
+                                                                const std::vector<int>& vregion)
+{
+    BOOST_OUTCOME_TRY(auto&& population_data, mio::read_population_data(path));
+    return read_population_data(population_data, vregion);
+}
+} // namespace mio
+#endif //MEMILIO_HAS_JSONCPP
--- a/cpp/memilio/io/parameters_io.h
+++ b/cpp/memilio/io/parameters_io.h
@@ -1,0 +1,143 @@
+/* 
+* Copyright (C) 2020-2025 MEmilio
+*
+* Authors: Henrik Zunker
+*
+* Contact: Martin J. Kuehn <Martin.Kuehn@DLR.de>
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+#ifndef MEMILIO_IO_PARAMETER_H
+#define MEMILIO_IO_PARAMETER_H
+
+#include "memilio/config.h"
+
+#ifdef MEMILIO_HAS_JSONCPP
+
+#include "memilio/io/epi_data.h"
+#include "memilio/io/result_io.h"
+#include "json/value.h"
+#include <string>
+#include <vector>
+
+namespace mio
+{
+/**
+ * @brief Gets the region ID (county, state, or district) of an EpiDataEntry. If none are available, 
+ * it defaults to 0 which is representing the whole country.
+ * 
+ * If none are available, it defaults to 0 which is representing the whole country.
+ * 
+ * @tparam EpiDataEntry The type of the data entry.
+ * @param data_entry The (RKI) data entry to extract the region ID from.
+ * @return The region ID as integer, or 0 if no specific region information is available.
+ */
+template <class EpiDataEntry>
+int get_region_id(const EpiDataEntry& data_entry)
+{
+    return data_entry.county_id ? data_entry.county_id->get()
+                                : (data_entry.state_id ? data_entry.state_id->get()
+                                                       : (data_entry.district_id ? data_entry.district_id->get() : 0));
+}
+
+/**
+ * @brief Extracts the number of individuals in critical condition (ICU) for each region 
+ * on a specified date from the provided DIVI data.
+ *
+ * @tparam FP Floating point type (default: double).
+ *
+ * @param[in] divi_data Vector of DIVI data entries containing date, region, and ICU information.
+ * @param[in] vregion Vector of region IDs for which the data is computed.
+ * @param[in] date Date for which the ICU data is computed.
+ * @param[in, out] vnum_icu Output vector containing the number of ICU cases for each region.
+ *
+ * @return An IOResult indicating success or failure.
+ */
+template <typename FP = double>
+IOResult<void> compute_divi_data(const std::vector<DiviEntry>& divi_data, const std::vector<int>& vregion, Date date,
+                                 std::vector<FP>& vnum_icu)
+{
+    auto max_date_entry = std::max_element(divi_data.begin(), divi_data.end(), [](auto&& a, auto&& b) {
+        return a.date < b.date;
+    });
+    if (max_date_entry == divi_data.end()) {
+        log_error("DIVI data is empty.");
+        return failure(StatusCode::InvalidValue, "DIVI data is empty.");
+    }
+    auto max_date = max_date_entry->date;
+    if (max_date < date) {
+        log_error("DIVI data does not contain the specified date.");
+        return failure(StatusCode::OutOfRange, "DIVI data does not contain the specified date.");
+    }
+
+    for (auto&& entry : divi_data) {
+        auto it      = std::find_if(vregion.begin(), vregion.end(), [&entry](auto r) {
+            return r == 0 || r == get_region_id(entry);
+        });
+        auto date_df = entry.date;
+        if (it != vregion.end() && date_df == date) {
+            auto region_idx      = size_t(it - vregion.begin());
+            vnum_icu[region_idx] = entry.num_icu;
+        }
+    }
+
+    return success();
+}
+
+/**
+ * @brief Reads DIVI data from a file and computes the ICU data for specified regions and date.
+ *
+ * @tparam FP Floating point type (default: double).
+ *
+ * @param[in] path Path to the file containing DIVI data.
+ * @param[in] vregion Vector of region IDs for which the data is computed.
+ * @param[in] date Date for which the ICU data is computed.
+ * @param[in, out] vnum_icu Output vector containing the number of ICU cases for each region.
+ *
+ * @return An IOResult indicating success or failure.
+ */
+template <typename FP = double>
+IOResult<void> read_divi_data(const std::string& path, const std::vector<int>& vregion, Date date,
+                              std::vector<FP>& vnum_icu)
+{
+    BOOST_OUTCOME_TRY(auto&& divi_data, mio::read_divi_data(path));
+    return compute_divi_data(divi_data, vregion, date, vnum_icu);
+}
+
+/**
+ * @brief Reads population data from a vector of population data entries.
+ * 
+ * @param[in] population_data Vector of population data entries.
+ * @param[in] vregion Vector of keys representing the regions of interest.
+ * @return An IOResult containing a vector of vectors, where each inner vector represents the population
+ *         distribution across age groups for a specific region, or an error if the function fails.
+ */
+IOResult<std::vector<std::vector<double>>> read_population_data(const std::vector<PopulationDataEntry>& population_data,
+                                                                const std::vector<int>& vregion);
+
+/**
+ * @brief Reads population data from census data.
+ * 
+ * @param[in] path Path to the population data file.
+ * @param[in] vregion Vector of keys representing the regions of interest.
+ * @return An IOResult containing a vector of vectors, where each inner vector represents the population
+ *         distribution across age groups for a specific region, or an error if the function fails.
+ */
+IOResult<std::vector<std::vector<double>>> read_population_data(const std::string& path,
+                                                                const std::vector<int>& vregion);
+
+} // namespace mio
+
+#endif //MEMILIO_HAS_JSONCPP
+
+#endif //MEMILIO_IO_PARAMETER_H
--- a/cpp/models/ode_secir/parameters_io.cpp
+++ b/cpp/models/ode_secir/parameters_io.cpp
@@ -30,6 +30,7 @@ GCC_CLANG_DIAGNOSTIC(ignored "-Wmaybe-uninitialized")
 
 #include "ode_secir/parameters_io.h"
 #include "memilio/io/epi_data.h"
+#include "memilio/io/parameters_io.h"
 #include "memilio/io/io.h"
 #include "memilio/utils/stl_util.h"
 #include "memilio/utils/date.h"
@@ -42,15 +43,6 @@ namespace osecir
 
 namespace details
 {
-//district, county or state id of a data entry if available, 0 (for whole country) otherwise
-//used to compare data entries to integer ids in STL algorithms
-template <class EpiDataEntry>
-int get_region_id(const EpiDataEntry& entry)
-{
-    return entry.county_id
-               ? entry.county_id->get()
-               : (entry.state_id ? entry.state_id->get() : (entry.district_id ? entry.district_id->get() : 0));
-}
 //overload for integers, so the comparison of data entry to integers is symmetric (required by e.g. equal_range)
 int get_region_id(int id)
 {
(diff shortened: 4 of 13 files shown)


cpp/README.md
```diff
@@ -33,6 +33,7 @@ The following table lists the dependencies that are used. Most of them are requi
 | JsonCpp | 1.9.6    | No       | Yes (git repo)        | https://github.com/open-source-parsers/jsoncpp |
 | HDF5    | 1.12.0   | No       | No                    | https://www.hdfgroup.org/, package libhdf5-dev on apt (Ubuntu) |
 | GoogleTest | 1.10  | For Tests only | Yes (git repo)  | https://github.com/google/googletest |
+| LibSBML | 5.18.0 | For SBML integration only | Yes (sourgeforce repo) |  https://sbml.org/software/libsbml/ |
```
HenrZu, line 36: i would prefer to fill Required with `No`. And also apply this to the GoogleTest part. Would make it more consistent with the libaries above. The current text could be put into the notes

cpp/examples/sbml2memilio.cpp
```diff
@@ -0,0 +1,53 @@
+#include "sbml2memilio.h"
```
HenrZu, line 1: we already have a lot of examples. Im against splitting this in 2 files.  As this is not part of the libary, this is fine in my opinion.
Also the naming is not alligning with our notation. Use underscores :)

cpp/examples/sbml2memilio.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+int main(int argc, char* argv[])
```
HenrZu, line 3: write a short summary, what this example is doing.

cpp/examples/sbml2memilio.h
```diff
@@ -0,0 +1,856 @@
...
+#include <fstream>
+#include <string>
+#include <vector>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/filesystem.hpp>
+
+#include <sbml/SBMLTypes.h>
+
+LIBSBML_CPP_NAMESPACE_USE
```
HenrZu, line 12: please comment on /explain this

cpp/examples/sbml2memilio.h
```diff
@@ -0,0 +1,856 @@
...
+
+/*
+    @brief Get the filename
+
+    @param filename The filename to be processed
+
+    Uses `boost::filesystem::path` to get the path to the input file, then uses `stem()` to get the name of the input file without file ending.
+*/
+std::string get_filename(const char* filename)
```
HenrZu, line 21: the doc for each function is nice. However, there could be more comment about the implication into the overall context.

cpp/examples/sbml2memilio.h
```diff
@@ -0,0 +1,856 @@
+#include <cmath>
```
HenrZu, line 1: Also, is this in not an example in my opinion. It just more a file to transform sbml models into memilio models. Dont think this should be part of the example dir. Maybe we should discuss where to place it.

cpp/examples/sbml2memilio.cpp
```diff
@@ -0,0 +1,53 @@
...
+{
+    if (argc != 2) {
+        std::cout << "Please provide a SBML file at startup!" << std::endl << std::endl;
+        return 1;
+    }
+    const char* filename = argv[1];
+    SBMLDocument* document;
```
HenrZu, line 10: do we need to use a raw pointer here?

cpp/examples/CMakeLists.txt
```diff
@@ -136,6 +136,12 @@ add_executable(smm_example smm.cpp)
 target_link_libraries(smm_example PRIVATE memilio smm)
 target_compile_options(smm_example PRIVATE ${MEMILIO_CXX_FLAGS_ENABLE_WARNING_ERRORS})
 
+if(MEMILIO_HAS_SBML)
+    add_executable(sbml2memilio sbml2memilio.cpp)
+    target_link_libraries(sbml2memilio PRIVATE sbml memilio) #ersetzen um libsbml zu finden
```
HenrZu, line 141:
```suggestion
    target_link_libraries(sbml2memilio PRIVATE sbml memilio)

```

cpp/thirdparty/CMakeLists.txt
```diff
@@ -196,6 +196,14 @@ else()
         to the directory containing the jsoncppConfig.cmake file to build with JsonCpp.")
 endif()
```
HenrZu, line 198: above we fix the required versions for some libs. Is this necessary for SBML?

cpp/examples/sbml2memilio.h
```diff
@@ -0,0 +1,856 @@
...
+#include <iostream>
+#include <fstream>
+#include <string>
+#include <vector>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/filesystem.hpp>
+
+#include <sbml/SBMLTypes.h>
+
```
HenrZu, line 11: add #ifdef MEMILIO_HAS_SBML here

docs/source/cpp/sbml.rst
```diff
@@ -0,0 +1,65 @@
...
+The `SBML <https://sbml.org/>`_ integration into memilio works via the ``sbml2memilio`` executable. 
+Building it requires the installation of `libsbml <https://sbml.org/software/libsbml/>`_ . Then the general build 
+command will produce the executable in the `build/bin`-directory. 
+
+Building the converter
+---------------------------------
+
+To build the executable, the following steps are necessary:
+
+1. Install ``libsbml``
```
HenrZu, line 13: write a bit more here. Maybe just "using the provided link above. Or move this part into this step here. At least provide the link

docs/source/cpp/sbml.rst
```diff
@@ -0,0 +1,65 @@
...
+2. Create the cmake files. In our tests, libsbml was never found by default. Therefore, you have to hint cmake to the location of the ``sbml-config.cmake``, e.g. by adding the flag ``-Dsbml_DIR=<path-to-libsbml>``.
```
HenrZu, line 14:
```suggestion
2. Create the Cmake files. In our tests, libsbml was never found by default. Therefore, you have to hint cmake to the location of the ``sbml-config.cmake``, e.g. by adding the flag ``-Dsbml_DIR=<path-to-libsbml>``.
```

docs/source/cpp/sbml.rst
```diff
@@ -0,0 +1,65 @@
...
+3. Build the project with the general build command.
+4. The output files will only be nicely formatted, if `clang-format <https://clang.llvm.org/docs/ClangFormat.html>`_ is installed. It will work also without it, but installing it is recommended.
+
+For more information on the build process, please refere to :doc:`../getting_started`.
```
HenrZu, line 18:
```suggestion
For more information on the build process, please refer to :doc:`../getting_started`.
```

docs/source/cpp/sbml.rst
```diff
@@ -0,0 +1,65 @@
...
+Limitations
+-------------
+Not every feature implemented in SBML is also supported by memilio.
+The following features are the most important not supported features:
+
+    - Events not triggered by time
+    - Usage of multiple SBML-compartments
+    - Rules (unless they are RateRules)
+
+As in an SBML file all species have to be given seperately, it is unfortunately also not possible to automatically use the Population in MEmeilio to stratify the population into different compartments.
```
HenrZu, line 58:
```suggestion
As in an SBML file all species have to be given seperately, it is unfortunately also not possible to automatically use the Population in MEmilio to stratify the population into different compartments.
```

HenrZu (changes requested):
Great addition! i had a first look at it and have some comments. I still need to look into the sbml.h file more in detail and also into the doc.

cpp/examples/sbml_to_memilio.cpp
```diff
@@ -0,0 +1,921 @@
...
+
+    std::string parameterset_initializer = "using ParametersBase = ParameterSet<";
+
+    if (number_parameters == 0) {
+        parameters << "template <typename FP = ScalarType>\nstruct DummyParam {\n    using Type = double;\n    static "
+                      "Type get_default()\n    {\n        return 0.0;\n    }\n    static std::string name()\n    {\n   "
+                      "     return \"DummyParameter\";\n    }\n};\n"
+                   << std::endl;
+        parameterset_initializer = parameterset_initializer + "DummyParam<FP>, ";
+    }
```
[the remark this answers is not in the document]
reneSchm: This should also allow you to remove the entire number_parameters==0 if-block (while keeping the loop from the else).

cpp/examples/sbml_to_memilio.cpp
```diff
@@ -0,0 +1,921 @@
...
+{
+
+    std::string lowercase_name = boost::to_lower_copy<std::string>(get_filename(filename));
+
+    std::ofstream model_cpp;
+    model_cpp.open(lowercase_name + "/model.cpp", std::ios::out);
+
+    model_cpp << "#include \"" << lowercase_name << "/model.h\"\n\nnamespace mio\n{" << std::endl;
+    model_cpp << "namespace " << lowercase_name << "\n{\n\n}\n}" << std::endl;
+    model_cpp.close();
```
reneSchm, line 461: Another note on file operations, since you use the file (via fstream) in a very limited scope, you don't *have to* call close manually, the destructor can take care of that. I am not sure what is best practice here, so I leave it up to you whether you want to keep the close, or remove them.

cpp/examples/sbml_to_memilio.cpp
```diff
@@ -0,0 +1,921 @@
...
+
+    Creates the model.h file based on the filename. First some generic input is written to the file. Then the `get_derivatives`-function is generated. It
+    - creates an index variable for every species
+    - creates a lambda function for every function definition in the model to have it at hand
+    - goes through every reaction in the model and generates the corresponding code. This is done in local scopes to make sure local parameters don't cause errors.
+    - checks every rule whether it is a RateRule and generates code for it if it is.
+    
+    In the end it appends a generic serialization function.
+*/
+bool create_model_h(Model* model, const char* filename)
```
reneSchm, line 480: There is a lot of nested logic in this function, can you maybe move some of this into helper functions? Can some of it be reused? Because as is, this function is really hard to read and understand.

Optimally, I'd like to see no more than two indents, and have them closed as soon as possible. But do not take that as a requirement, only an aspiration.

cpp/examples/sbml_to_memilio.cpp
```diff
@@ -0,0 +1,921 @@
...
+    const char* filename = argv[1];
+    SBMLDocument* document;
+    SBMLReader reader;
+
+    document = reader.readSBML(filename);
+
+    auto model = document->getModel();
+
+    if (!verify_model_suitability(model)) {
+        return 1;
```
reneSchm, line 885: Most commonly used exit codes start at 127, so if you like, you could use incrementing exit codes so it is easier to see which function failed.

cpp/examples/sbml_to_memilio.cpp
```diff
@@ -0,0 +1,921 @@
...
+            parameters << "template <typename FP = ScalarType>" << std::endl;
+            parameters << "struct " << name << " {\n    using Type = double;\n  static Type get_default()\n{"
+                       << std::endl;
+            parameters << "return " << value << ";\n}" << std::endl;
+            parameters << "static std::string name()\n{\n   return \"" << name << "\";\n}\n};\n" << std::endl;
+            parameterset_initializer = parameterset_initializer + name + "<FP>, ";
+        }
+    }
+    parameterset_initializer.pop_back();
+    parameterset_initializer.pop_back();
```
reneSchm, line 429: I would prefer rewriting the loop logic, maybe by adding something like
```
if (i != 0) { parameterset_initializer += ", "; }
```
at the start of the loop.

cpp/examples/sbml2memilio.cpp
reneSchm: Looking at SBMLReader.cpp, the pointer returned by readSBML is owning and should therefore be accepted as a uniqe_ptr:
```
std::unique_ptr<SBMLDocument> document(reader.readSBML(filename));
```

The Model*, however, is non-owning, so we should accept it as a (const, if possible) Model&, i.e.
```
Model& model = *document->getModel();
```
The function arguments should be changed accordingly.

reneSchm: Also, the library uses class, virtual, and std::string, it is very much C++ (or has a C++ interface). But anyway, we should not use raw pointers, if avoidable.

cpp/examples/sbml_to_memilio.cpp
```diff
@@ -0,0 +1,921 @@
...
+#include <string>
+#include <vector>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/filesystem.hpp>
+
+#include "memilio/config_internal.h"
+
+#ifdef MEMILIO_HAS_SBML
+#include <sbml/SBMLTypes.h>
```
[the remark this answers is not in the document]
reneSchm: Also, is the `#ifdef MEMILIO_HAS_SBML` needed, when the entire file is guarded in the cmake?

docs/source/cpp/sbml.rst
```diff
@@ -0,0 +1,65 @@
...
+Building it requires the installation of `libsbml <https://sbml.org/software/libsbml/>`_ . Then the general build 
+command will produce the executable in the `build/bin`-directory. 
+
+Building the converter
+---------------------------------
+
+To build the executable, the following steps are necessary:
+
+1. Install ``libsbml``. The latest releases are available on `GitHub <https://github.com/sbmlteam/libsbml/releases>`_, advice can be found in the `documentation <https://sbml.org/software/libsbml/libsbml-docs/api/c/>`_.
+2. Create the Cmake files. In our tests, libsbml was never found by default. Therefore, you have to hint cmake to the location of the ``sbml-config.cmake``, e.g. by adding the flag ``-Dsbml_DIR=<path-to-libsbml>``.
```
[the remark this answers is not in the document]
reneSchm: Thats on me, github likes to view angle brackets as html, which hid the important part of the comment. Edited above

cpp/examples/sbml_to_memilio.cpp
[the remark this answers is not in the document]
reneSchm: I see, but a that point the includes were still in a header file. There, the ifdef made sense. Here, not as much.

cpp/examples/sbml_to_memilio.cpp
```diff
@@ -0,0 +1,921 @@
...
+
+/*
+    @brief Returns the maximal time mentioned in a formula
+
+    @param trigger The formula to be processed as ASTNode*
+    @param model The model where the formula stems from
+    @param name_namespace The namespace of the model
+
+    This function assumes a very specific layout of the formula. It needs to be a comparison between exaclty two values. If that is not the case, it throws an error. Then it expects one of the two noeds to be a time node. The other node has the comparison value and is returned. It is always assumed, but never checked, that we have a positive comparison, i.e. the time is indeed the maximum time.
+    This function is used to find the maximum time until which an event runs. It is used for the generation of the example.cpp file.
```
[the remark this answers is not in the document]
reneSchm: > This is not really error handling, it just returns a different string for which the next function searches?

It definitely is error handling, at least it is framed as such. The search literally checks `tmax == "Error"`, and your docstring says "throws an error". It would be better to return an IOResult, the workflow would stay mostly the same, but you can immediately parse tmax, or say that you can't. It also makes it obvious from the function signature alone, that it can fail.

On a side note, if your function does not throw exceptions, avoid the word throw, as it is a C++ keyword. And since we do not use exceptions in memilio, I guess I am saying: please don't throw anything.

docs/source/cpp/sbml.rst
```diff
@@ -0,0 +1,65 @@
...
+Usage
+------
+Call the executable on a given SBML file as 
+
+.. code:: bash
+
+    ./sbml2memilio <sbml-file>
+
+
+If `clang-format <https://clang.llvm.org/docs/ClangFormat.html>`_ is not installed, it will end with an error, but produce the necessary files nevertheless.
```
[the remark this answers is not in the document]
reneSchm: Oh, good point. Maybe there is an option to disable that behaviour, otherwise we could add a log of our own? Though I am not sure what loglevel to pick, maybe a printf to inform that this (error) is fine would be ok here.

cpp/examples/sbml_to_memilio.cpp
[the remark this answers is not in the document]
reneSchm: Add the line `template <class Dummy = void, class = std::enable_if_t<(sizeof...(Tags)), Dummy>, class = Dummy>` below "private:" in parameter_set.h:344.

This fixes the default constructor for specifically ParameterSet\<> (without parameters), as the private constructor in parameter_set.h:345 was incorrectly used, due to it being a better "fit".

With this one line you no longer need to handle the extra case "number_parameters==0", adding a single bogus parameter, anywhere, since an empty ParameterSet is perfectly valid (though useless).

cpp/memilio/utils/parameter_set.h
```diff
@@ -342,6 +342,9 @@ class ParameterSet
     }
 
 private:
+    // Fix default constructor for empty parameter set
+    template <class Dummy = void, class = std::enable_if_t<(sizeof...(Tags)), Dummy>, class = Dummy>
+
     ParameterSet(const typename Tags::Type&... t)
```
reneSchm, line 348: You are right, this template should be explained.
```suggestion
    /**
     * @brief Internal constructor taking a value for each parameter, and initializing it.
     * @tparam Dummy Do not specify template parameters. They are used to select a viable constructor. The second and
     * third parameter are needed for an empty ParameterSet<>, to disable this constructor and avoid equivalence to
     * other constructors, respectively.
     */
    template <class Dummy = void, class = std::enable_if_t<(sizeof...(Tags)), Dummy>, class = Dummy>
    ParameterSet(const typename Tags::Type&... t)
```

cpp/sbml_model_generation/sbml_to_memilio.cpp
```diff
@@ -0,0 +1,1053 @@
...
+
+/**
+ * @brief Create the additional lines for the CmakeLists.txt file
+ *
+ * @param filename The name of the file to be processed
+ * @param path The path where the file should be created
+ *
+ * Appends the necessary commands for compilation to the CMakeLists.txt file in path using the filename.
+*/
+bool modify_cmakelists(const std::string& filename, const std::string& path)
```
reneSchm, line 930: Does this work when run repeatedly on the same file?

Maybe you could do search for the line "add_subdirectory(lowercase_name)" before appending it again?

cpp/sbml_model_generation/sbml_to_memilio.cpp
```diff
@@ -0,0 +1,1053 @@
...
+        return 11;
+    }
+
+    if (!modify_cmakelists(core_filename, path)) {
+        return 12;
+    }
+    mio::log_info("Created all files.");
+    mio::log_info("Formatting files.");
+    format_files(core_filename, path);
+}
```
reneSchm, line 1053:
```suggestion
}

```

cpp/sbml_model_generation/sbml_to_memilio.cpp
```diff
@@ -0,0 +1,1053 @@
...
+    if (trailing_bracket.size() > 0) {
+        formula = formula + trailing_bracket;
+    }
+}
+
+/**
+ * @brief Formats a formula for the example.cpp file
+ *
+ * @param model The model where the formula stems from
+ * @param math_string The formula to be processed, preprocessed as char*
```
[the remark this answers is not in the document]
reneSchm: Okay, if I hadn't read the code, I would not know what you are talking about at all.
So please add some context there. Something like "[...] preprocessed from x into y", or "A formula from the AST/SBML model in string representation"?

cpp/sbml_model_generation/sbml_to_memilio.cpp
```diff
@@ -0,0 +1,1093 @@
...
+ * @param[in] model_namespace The namespace of the model.
+ * @return st::string The formatted initial assignement.
+ *
+ * First it checks whether an initial amount is given for the species. Then it checks the other possibilites for 
+ * initial assignements:
+ * - If an initial concentration is given, it is used
+ * - If an initial assignment is given, it is used. This could be
+ *   - a values
+ *   - a other parameter
+ *   - a formula, which would then be parsed using :cpp:func:`format_main_formula(Model * model, char* math_string, std::string * model_namespace)`
```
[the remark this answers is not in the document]
reneSchm: I am slightly concerned about mixing rst into the doxygen documentation, maybe that is a topic for the next meeting.

Doxygen does link classes and functions usually by name only. There is also the `@see` keyword, see [DampingMatrixExpression](https://memilio.readthedocs.io/en/latest/api/classmio_1_1DampingMatrixExpression.html#exhale-class-classmio-1-1dampingmatrixexpression) for a (somewhat random) example with both present.

.github/actions/sbml-test/action.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - name: Run sbml importer
+      shell: bash
+      run: | 
+        cd cpp
+        ./build/bin/sbml_to_memilio tests/data/sbml_test.sbml
+    - name: Compile result
+      shell: bash
+      run: |
+        cd cpp/build
+        cmake -DCMAKE_BUILD_TYPE=${{ inputs.config }} -DMEMILIO_ENABLE_IPOPT=ON -Dsbml_DIR=/usr/lib/x86_64-linux-gnu/cmake ..
```
reneSchm, line 64: Input does not exist. Also why enable IPOPT?

cpp/tests/data/sbml_test.sbml
```diff
@@ -0,0 +1,330 @@
...
+                    <cn> 0.2 </cn>
+                    <cn> 0.05 </cn>
+                  </apply>
+                  <pi/>
+                </apply>
+                <ci> Comp1 </ci>
+              </apply>
+              <apply>
+                <times/>
+                <csymbol encoding="text" definitionURL="http://www.sbml.org/sbml/symbols/time"> time </csymbol>
```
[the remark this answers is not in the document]
reneSchm: Weird, but fine, I guess.

.github/actions/sbml-test/action.yaml
```diff
@@ -0,0 +1,65 @@
...
+        sudo apt-get -qq -y install gcc-13
+        sudo apt-get -qq -y install g++-13
+        sudo apt-get -qq -y install libbz2-dev libbz2-1.0
+        sudo apt-get -qq -y install libsbml5t64 libsbml5-dev
+        sudo apt-get -qq -y install ccache
+        sudo apt-get -qq update
+    - name: ccache
+      id: ccache
+      uses: hendrikmuhs/ccache-action@v1.2
+      with:
```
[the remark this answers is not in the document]
reneSchm: Hm, it does reuse a good bit of the old build, but recompiles some of memilio... I think maybe HDF5 is the issue, can you check that all deps used by the gcc-latest build are also installed here? (See apt commands)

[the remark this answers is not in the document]
HenrZu: I vaguely remember that I had the same problem with the models in a different setup. However, didnt find a solution at that time.
It's not ideal but we can also outsource this to an issue and accept it for now.

reneSchm (approved):
Please merge main again so that the CI can pass, then I think this is good to go.

**Note**: This new functionality is not covered by unit tests, instead it has its own CI job.

HenrZu (approved):

PR #1414, opened 2025-11-11
Improve messages related to parameter checks

### LINKED ISSUE #1413: Inconsistent error messages in several models
### Bug description

In (at least) the `sde_sir`, `sde_sirs`, `ode_seir`, `ode_seirvv`, `ode_sir` models, the `check_constraints` function compares some values to a `tol_times` parameter, but then prints a warning that the parameter value is smaller than 0. Also, this should be warnings instead of errors.

### Version

Any

### To reproduce

Run any of the above simulations with all parameters set to 0.05.

### Relevant log output

```zsh
[2025-11-10 16:43:40.072] [error] Constraint check: Parameter TimeInfected 0.0427 smaller or equal 0.0000. Please note that unreasonably small compartment stays lead to massively increased run time. Consider to cancel and reset parameters.
```

```zsh

```

### Checklist

- [x] Linked to project

kilianvolmer:
# Changes and Information

- When parameters are compared to a tolerance, the message on failure prints the tolerance
- The formatting of values in messages is removed to reduce the possibility of wrong formatting
- For populations, we now return errors if values are smaller than -1e-10, otherwise warnings if the value is set to zero in apply_constraints().

-
-

### Checks by code author

- [x] New code adheres to [coding guidelines](https://memilio.readthedocs.io/en/latest/development.html#coding-guidelines)
- [x] Appropriate **documentation within the code** (Doxygen) for new functionality has been added in the code
- [x] Appropriate **external documentation** (ReadTheDocs) for new functionality has been added to the online documentation

Closes #1413 #1415 #1416

### DIFF
--- a/cpp/models/ode_mseirs4/parameters.h
+++ b/cpp/models/ode_mseirs4/parameters.h
@@ -221,10 +221,12 @@ public:
 
     bool check_constraints() const
     {
+        bool violated = false;
+
         auto check_nonneg = [&](auto& v) {
             if (v < 0) {
                 log_error("Constraint check: Parameter was found to be {} and should not be smaller than {}.", v, 0);
-                return true;
+                violated = true;
             }
         };
         check_nonneg(this->template get<BaseTransmissionRate<FP>>());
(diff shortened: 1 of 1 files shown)


mknaranja:
For populations, we now return errors if values are smaller than -1e-10, otherwise warnings if the value is set to zero in apply_constraints(). After some consideration, I am unsure if we want to blow up the code similarly for all models. I'd rather go with no. Maybe a slightly bigger refactoring with lambda functions such as in the RSV model might be more suited anyway and would pave the base for these additional changes.

reneSchm:
But that differentiation does not matter - the "do and tell" part is important. We react to both cases in the exact same way (i.e. clamping to 0), and give the exact same log message. So you are just hoping that someone has the same associations with the log levels that you made.

As a general guideline, the log level is not part of the message you want to send, just a way to categorize (and filter by) importance.

Hence, the bare minimum would be change the message for each case. But as both cases have the same end-result, that seems unnecessary. Thus my proposal of always using error for higher visibility. The message already contains new and old value, so the user can already see how bad the correction was.

Maybe what you want to do is change how the apply_constraint result is handled in your code? I don't know, you have not given me much context to work with.

PR #1475, opened 2026-01-12
Improve compile time of Flows

reneSchm:
# Changes and Information

A fairly small change to the index_of_type implementation results in a significant reduction in time and memory used during compilation of larger flow models. Tested on the off-repo sde_ui10 that uses ~400 flows, resulting in a speedup of 4 while using 1/5 of the memory. This is a rough measurement using `/usr/bin/time` on clean compiling the ui10 example, [see the attached file](https://github.com/user-attachments/files/24570346/timing-results.txt) for details. 

## Merge Request - Guideline Checklist

Please check our [git workflow](https://memilio.readthedocs.io/en/latest/development.html#git-workflow). Use the **draft** feature if the Pull Request is not yet ready to review.

### Checks by code author

- [ ] Every addressed issue is linked (use the "Closes #ISSUE" keyword below)
- [x] New code adheres to [coding guidelines](https://memilio.readthedocs.io/en/latest/development.html#coding-guidelines)
- [x] No large data files have been added (files should in sum not exceed 100 KB, avoid PDFs, Word docs, etc.)
- [x] Tests are added for new functionality and a local test run was successful (with and without OpenMP)
- [x] Appropriate **documentation within the code** (Doxygen) for new functionality has been added in the code
- [x] Appropriate **external documentation** (ReadTheDocs) for new functionality has been added to the online documentation
- [x] Proper attention to licenses, especially no new third-party software with conflicting license has been added
- [ ] (For ABM development) Checked [benchmark results](https://memilio.readthedocs.io/en/latest/development.html#agent-based-model-development) and ran and posted a local test above from before and after development to ensure performance is monitored.

### Checks by code reviewer(s)

- [ ] Corresponding issue(s) is/are linked and addressed
- [x] Code is clean of development artifacts (no deactivated or commented code lines, no debugging printouts, etc.)
- [x] Appropriate **unit tests** have been added, CI passes, code coverage and performance is acceptable (did not decrease)
- [x] No large data files added in the whole history of commits(files should in sum not exceed 100 KB, avoid PDFs, Word docs, etc.)
- [ ] On merge, add 2-5 lines with the changes (main added features, changed items, or corrected bugs) to the merge-commit-message. This can be taken from the **briefly-list-the-changes** above (best case) or the separate commit messages (worst case).

PR #1476, opened 2026-01-14
Improve compile time of Flows

reneSchm:
This PR works around an apparent github bug in #1475. Look at #1475 for the review of this code.


HenrZu (approved):
Already approved in bugged PR (see description in this PR)

PR #1535, opened 2026-04-07
1027 Use FetchContent_MakeAvailable instead of FetchContent_Populate

### LINKED ISSUE #1027: Use FetchContent_MakeAvailable instead of FetchContent_Populate?

Currently we use the structure 

```
    FetchContent_GetProperties(eigen)
    if(NOT eigen_POPULATED)
        FetchContent_Populate(eigen)
    endif()
```

for most dependencies. In some cases, it might be possible to use `FetchContent_MakeAvailable()` which is recommended (see also
https://cmake.org/cmake/help/latest/module/FetchContent.html) instead of using `FetchContent_GetProperties()` and `FetchContent_Populate()` directly.

### Enhancement description

Use `FetchContent_MakeAvailable()` instead of `FetchContent_GetProperties()` and `FetchContent_Populate()`. We may need to set the option OVERRIDE_FIND_PACKAGE or similar to prevent the search for local versions.

We could also discuss whether we want to allow searching for local versions and only downloading otherwise (and not controlling this via the MEMILIO_USE_BUNDLED_ options). This could speed up the build especially because of downloading boost.

### Additional context

_No response_

### Checklist

- [ ] Linked to project

reneSchm:
# Changes and Information

- Replace the deprecated FetchContent_Populate.
- Increase CMake minimum version to 3.18.
- Remove unused or outdated CMake files and code.
- Make config path in benchmarks independent of the execution directory

- Automatic fallbacks from FetchContent to find_package (or vice versa) via OVERRIDE_FIND_PACKAGE or FIND_PACKAGE_ARGS are only supported since cmake 3.24, which is newer than the most recent version in Ubuntu 22 (3.22)
- We may want to keep 1027 open if we want to reconsider automatic fallbacks in the future. Although I find the current solution is fine.

### Checks by code author

Closes #1027


cpp/tests/cmake/AddGoogleTest.cmake
```diff
@@ -6,35 +6,12 @@
 #
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
```
HenrZu, line 7: my format on save is making a correction regarding whiteapaces in l31.

cpp/benchmarks/CMakeLists.txt
```diff
@@ -5,18 +5,13 @@ set(BENCHMARK_ENABLE_INSTALL OFF CACHE BOOL "Don't install benchmark" FORCE)
 set(BENCHMARK_DOWNLOAD_DEPENDENCIES OFF CACHE BOOL "Don't download dependencies" FORCE)
 set(BENCHMARK_ENABLE_GTEST_TESTS OFF CACHE BOOL "Disable Google Test in benchmark" FORCE)
 
+message(STATUS "Downloading GoogleBenchmark library")
 include(FetchContent)
 FetchContent_Declare(benchmark
+    SOURCE_SUBDIR this_directory_name_does_not_exist
```
HenrZu:
```suggestion
```

HenrZu (changes requested):
Great update :)
There is a bug in the benchmark CMakeLists
`FetchContent_MakeAvailable` loads the code and then tries to call `add_subdirectory `to create thje target. However, with the path, CMake isnt able to find the CMakeList and the add_subdirectory is skipped. This is only working for Eigen (which is header-only)
Removing the line 11 solves this.

cpp/benchmarks/CMakeLists.txt
reneSchm: Good catch, that flag should have been removed. I had used the same workaround as for Eigen here, as the old code used to add an EXCLUDE_FROM_ALL, which MakeAvailable only supports from cmake 3.24 onward.

But in testing I noticed that exclude is not really necessary, as the add_subdirectory is gated behind the MEMILIO_BUILD flag, (same for benchmark).

cpp/tests/cmake/AddGoogleTest.cmake
reneSchm: Probably trailing white space. Strange, I thought I had autoformatting for CMake files working

HenrZu (approved):

PR #1544, opened 2026-04-17
1429 move example output to example_results

### LINKED ISSUE #1429: Move example output to example_results

Some examples create output files in the directory they are executed in. When using VSCode, this is the build/bin directory, making the files hard to find as well as cluttering the build directory with unrelated files.

### Enhancement description

Change examples so that they create files inside `memilio/example_results/`, by using e.g. 

```cpp
const std::string result_dir = mio::path_join(mio::base_dir(), "example_results");
```

This way the output of the examples is predictable, and more visible. We may want to add that path to the .gitignore.

### Additional context

The examples ode_secir_parameter_study.cpp and abm_parameter_study.cpp already do this.

### Checklist

- [x] Linked to project

### LINKED ISSUE #1557: Use Python 3.9 instead of 3.8 in CI
### Bug description

The CI fails because Python 3.8 is not supported from pypa/manylinux anymore, see https://github.com/pypa/manylinux/commit/85c2bf2350fd785c5c83d8861e7d548ef86634f4.

### Version

Windows

### To reproduce

Run CI

### Relevant log output

```zsh

```

```zsh

```

### Checklist

- [x] Linked to project

reneSchm:
# Changes and Information

- Update implementation of base_dir() and data_dir() (formerly DATA_DIR).
  - Move to io/directories.h.
  - Make both customizable through CMake during configure step.
- Add a new example_results_dir() function, to provide a consistent place for files created by examples.
- Add create_directories_or_exit() as a helper for opening directories (with predictable behavior on errors).
- Use example_results_dir() (and create_directory_or_exit()) to place all files from current examples into example_results/.
- Simplify some examples by using print_table().
- Replace usage of boost::filesystem with std::filesystem.
- Remove namespace aliases from header files.
  - Move mio::GraphABModel to mio::abm::GraphABModel.
- Use std::filesystem::path for select path arguments instead of std::string.
- Make std::filesystem::path loggable.
- Update spdlog dependency to 1.17 to avoid compile errors with some const strings. 
- Speed up CI for SBML by not building the test suite.

- Sorry for the big changelog, the "cleanup" phase escalated slightly.
- It sure would be nice if M$ had not ignored the part of the standard saying `Paths are implicitly convertible to and from std::basic_strings`.

### Checks by code author


cpp/memilio/io/directories.h
```diff
@@ -17,11 +17,12 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-#ifndef MIO_BASE_DIR_H
-#define MIO_BASE_DIR_H
+#ifndef MIO_UTILS_DIRECTORIES_H
+#define MIO_UTILS_DIRECTORIES_H
```
HenrZu: This in no longer in the Utils dir.
```suggestion
#ifndef MIO_IO_DIRECTORIES_H
#define MIO_IO_DIRECTORIES_H
```

cpp/memilio/io/directories.h
```diff
@@ -30,11 +31,29 @@ namespace mio
...
+{
+    return details::MEMILIO_DATA_DIR;
+}
+
+/**
+ * @brief Returns the absolute path to a common ouput directory for the code examples.
+ */
+[[maybe_unused]] const static std::filesystem::path example_results_dir(const std::string& example_name)
+{
+    // the last empty string is used to end the output path in a /
```
HenrZu: Is this comment outdated?

cpp/memilio/io/directories.h
```diff
@@ -30,11 +31,29 @@ namespace mio
...
+    const static std::filesystem::path dir = base_dir() / "example_results" / example_name;
```
HenrZu: Why store dir as a local static var? So, the first example_name is reused for all later calls?

cpp/examples/ode_secir_parameter_study_graph.cpp
```diff
@@ -314,18 +315,15 @@ int main()
             ensemble_edges.emplace_back(std::move(std::get<2>(run)));
```
HenrZu, line 315: can you have a look a line 41 in this file? Here, we create` .../example_results/ode_secir_parameter_study` with `create_directory(..., abs_path)` without parents. If `example_results` doesnt exist yet, then this fails. 

I think it would be enough to add a true as argument for mio::create_directory

cpp/memilio/io/directories.h
```diff
@@ -30,11 +31,29 @@ namespace mio
 /**
  * @brief Returns the absolute path to the project directory.
  */
-const static std::string base_dir()
+const static std::filesystem::path base_dir()
```
HenrZu: THis might have an impact for others. Do we need to adjust the doc? i will also check this later. Please double check :)

cpp/examples/abm_minimal.cpp
```diff
@@ -165,10 +166,11 @@ int main()
     // The first column is Time. The other columns correspond to the number of people with a certain infection state at this Time:
```
HenrZu: see line 165. This is outdated. 

should be

`// The results are written into the file "example_results/abm_minimal/history.txt"`

HenrZu: docs/source/cpp/mobility_based_abm.rst
Line 347 -> Please update this also

HenrZu (changes requested):
Great addition and cleanup! Most important thing is the static dir, which is confusing to me. Otherwise minor comments .

cpp/memilio/io/directories.h
reneSchm: No, the static actually no longer makes sense here.
I started by just dumping every file directly into example_results, thus the path was static. But I thought it would be nicer to have one subdirectory for each example, esp. since some might write out entire graphs

cpp/memilio/io/directories.h
reneSchm: Good point, we should put out a message on merge. The docs do not mention or use base_dir or data_dir, and we have no generic page on writing examples to mention this in. So no updates required there.

Maybe we could add a "how to write examples page" or sth. to mention this in. But for now, I think this is sufficiently self-documented by usage (similar to path_join or the old create_directory).

cpp/examples/abm_minimal.cpp
reneSchm: Wow, good catch on the documentation. It still uses the old print_table order as well.

cpp/examples/ode_secir_parameter_study_graph.cpp
reneSchm: Wrong file, but correct remark ;)
Fixed.

reneSchm:
[tensorflow](https://pypi.org/project/tensorflow/) does not have a version that supports 3.14 on PyPi, so we probably stay on 3.13 for now

HenrZu (approved):

jubicker (approved):

PR #1552, opened 2026-04-27
1551 Build and test on MacOS

### LINKED ISSUE #1551: Build and test on MacOS

_No response_

### Enhancement description

Add CI jobs for building and testing on a recent MacOS version.

### Additional context

_No response_

### Checklist

- [x] Linked to project

reneSchm:
# Changes and Information

- Add build, test and run_example jobs to CI for Apple Clang on MacOS 26 image. 
- Bump most actions.
- In Windows CI, Download a HDF5 release instead of building from source.
- Reduce output of some examples.

- One tolerance was decreased in a newer test, as Apple Clang on M1 has different floating point behaviour from gcc/clang.
- pre-commit has no new version, needs to be updated later.
- Should we mention (limited) MacOS support now?

### Checks by code author

- [ ] Appropriate **external documentation** (ReadTheDocs) for new functionality has been added to the online documentation and checked in the preview.
- [x] (For ABM development) Checked [benchmark results](https://memilio.readthedocs.io/en/latest/development.html#agent-based-model-development) and ran and posted a local test above from before and after development to ensure performance is monitored.


cpp/examples/ide_initialization.cpp
```diff
@@ -125,8 +126,8 @@ int main(int argc, char** argv)
     sim.advance(tmax);
 
     // Print results.
-    sim.get_transitions().print_table({"S->E", "E->C", "C->I", "C->R", "I->H", "I->R", "H->U", "H->R", "U->D", "U->R"},
-                                      16, 8);
+    mio::interpolate_simulation_result(sim.get_transitions())
```
jubicker, line 129: Why did you add the interpolation?

cpp/models/abm/model.h
```diff
@@ -685,7 +685,7 @@ class Model
     int m_id; ///< Model id. Is only used for abm graph model or hybrid model.
     std::vector<Person> m_persons; ///< Vector of every Person.
     std::vector<Location> m_locations; ///< Vector of every Location.
-    std::vector<bool>
+    std::vector<char>
```
jubicker: Why did you change this? And does the initialization and functionality still work with char instead of bool?

cpp/examples/ide_initialization.cpp
reneSchm: I had to look at a lot of CI output, including examples, and this one added ~1k lines of output (dt=.01 for t in [-7, 2]). Now that is reduced to ~10.

cpp/models/abm/model.h
reneSchm: I meant to add a comment on this, but basically, vector\<bool\> is not a vector of booleans, but a specialized type with several differences to the generic vector. This is often considered a bug in the standard, and clang and gcc (partially) "fixed" this. For some reason, Apple Clang has not.

The result is that vector\<bool\> is not serializable, and a workaround is not portable (not even between some compiler versions). Also, one of the differences states:

> [Does not guarantee that different elements in the same container can be modified concurrently by different threads.](https://en.cppreference.com/cpp/container/vector_bool)

In particular, writing into an entry of vector\<bool\> may overwrite the entire vector. So it's maximally thread-unsafe.

vector\<char\> should work fine as a replacement, as chars can represent bools without issues. 
<details><summary> Another version I came up with would be to define (dropdown) </summary>

```cpp

namespace details
{
struct Boolean {
    Boolean() = default;
    Boolean(bool b)
        : value(b)
    {
    }
    bool operator==(bool other) const
    {
        return value == other;
    }
    bool operator!=(bool other) const
    {
        return value != other;
    }
    operator bool&()
    {
        return value;
    }
    operator bool() const
    {
        return value;
    }

    bool value;
};
} // namespace details

class VectorBool : public std::vector<details::Boolean>
{
    using std::vector<details::Boolean>::vector;
};

```
</details>

jubicker (approved):

PR #1574, opened 2026-06-01
1570 Allow setting tolerance for last timestep in SystemIntegrator manually

### LINKED ISSUE #1570: Allow setting tolerance for last timestep in SystemIntegrator manually

In SystemIntegrator::advance, the value `eps`  is used to determine the termination criterion of the time stepping scheme. With this it is ensured that the scheme reaches tmax up to a tolerance. The value of `eps` is set to the zero tolerance. In time stepping schemes with fixed time step size with a small time step size, this can lead to an additional time step at the end and thus to higher total number of time steps than expected. Setting `eps` to a larger value resolves this issue. 

### Enhancement description

Keep the default as is but allow changing `eps` manually. 

### Additional context

_No response_

### Checklist

- [x] Linked to project

annawendler:
# Changes and Information

- Add setter and getter for last_step_tolerance that determines if an additional time step is made to reach tmax

### Checks by code author

- [ ] Code is clean of development artifacts (no deactivated or commented code lines, no debugging printouts, etc.).
- [ ] Appropriate **unit tests** have been added, CI passes, code coverage and performance is acceptable (did not decrease).

Closes #1570

### DIFF
--- a/cpp/memilio/compartments/simulation_base.h
+++ b/cpp/memilio/compartments/simulation_base.h
@@ -154,6 +154,31 @@ public:
     }
     /** @} */
 
+    /**
+     * @brief Change the tolerance for the last time step.
+     * @param last_step_tolerance The new last_step_tolerance.
+     */
+    void set_last_step_tolerance(FP new_tol)
+    {
+        m_integrator.set_last_step_tolerance(new_tol);
+    }
+
+    /**
+     * @brief Access the tolerance for the last time step.
+     * @return A reference to the last_step_tolerance.
+     * @{
+     */
+    FP& get_last_step_tolerance()
+    {
+        return m_integrator.get_last_step_tolerance();
+    }
+
+    const FP& get_last_step_tolerance() const
+    {
+        return m_integrator.get_last_step_tolerance();
+    }
+    /** @} */
+
 protected:
     /**
      * @brief Run the simulation up to a given time.
--- a/cpp/memilio/math/integrator.h
+++ b/cpp/memilio/math/integrator.h
@@ -60,7 +60,7 @@ public:
     {
     }
 
-    virtual ~IntegratorCore() {};
+    virtual ~IntegratorCore(){};
 
     virtual std::unique_ptr<IntegratorCore<FP, Integrands...>> clone() const = 0;
 
@@ -188,8 +188,7 @@ public:
         using std::fabs;
         using std::max;
         using std::min;
-        const FP t0  = results.get_last_time();
-        const FP eps = Limits<FP>::zero_tolerance();
+        const FP t0 = results.get_last_time();
         assert(tmax > t0);
         assert(dt > 0);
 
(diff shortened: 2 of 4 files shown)


cpp/memilio/compartments/simulation_base.h
```diff
@@ -154,6 +154,31 @@ class SimulationBase
...
+
+    /**
+     * @brief Access the tolerance for the last time step.
+     * @return A reference to the last_step_tolerance.
+     * @{
+     */
+    FP& get_last_step_tolerance()
+    {
+        return m_integrator.get_last_step_tolerance();
+    }
```
reneSchm, line 174: This mashes two design philosophies. Either make a read-only getter (i.e. `const FP&` or simply `FP`) and a setter, or a read-write getter (without a setter). For integral types, which we can treat FP as, a read-write should be enough, as the setter won't do much.

cpp/memilio/math/integrator.h
```diff
@@ -271,10 +271,37 @@ class SystemIntegrator
...
+    FP& get_last_step_tolerance()
+    {
+        return m_last_step_tolerance;
+    }
+
+    const FP& get_last_step_tolerance() const
+    {
+        return m_last_step_tolerance;
+    }
+    /** @} */
```
reneSchm, line 299: See above

cpp/tests/test_compartments_simulation.cpp
```diff
@@ -90,6 +90,18 @@ TEST(TestCompartmentSimulation, copy_simulation)
...
+TEST(TestCompartmentSimulation, last_step_tolerance)
+{
+    auto sim = mio::Simulation<double, MockModel>(MockModel(), 0.0);
+
+    // Check default.
+    EXPECT_EQ(sim.get_last_step_tolerance(), mio::Limits<double>::zero_tolerance());
+    // Check setter of tolerance.
+    auto new_tol = 1e-2;
+    sim.set_last_step_tolerance(new_tol);
+    EXPECT_EQ(sim.get_last_step_tolerance(), new_tol);
```
reneSchm, line 102: A more meaningful test would be to reproduce the issue you want to solve here.

A simplified set up should work too, e.g. make the simulation take a 0.8 step until 1.0, once with a 0.25 tolerance, and once with the default. Such that the first one stops at 0.8, the second at 1.0.

cpp/tests/test_numericalIntegration.cpp
```diff
@@ -406,6 +406,20 @@ TEST(TestOdeIntegrator, integratorForcesLastStepSize)
     }
 }
 
+TEST(TestOdeIntegrator, integratorLastStepTolerance)
```
reneSchm, line 409: I think running essentially the same test twice has little benefit, right? The other test should cover this sufficiently

cpp/memilio/math/integrator.h
```diff
@@ -205,7 +204,8 @@ class SystemIntegrator
         FP dt_min_restore = m_core->get_dt_min(); // used to restore dt_min, if it was decreased to reach tmax
         FP t              = t0;
 
-        for (size_t i = results.get_num_time_points() - 1; fabs((tmax - t) / (tmax - t0)) > eps; ++i) {
+        for (size_t i = results.get_num_time_points() - 1; fabs((tmax - t) / (tmax - t0)) > m_last_step_tolerance;
```
reneSchm, line 207: We have `(tmax - t) / (tmax - t0) <= 1` and `(tmax - t) / (tmax - t0) --> 0`.
So we should add a requirement that `m_last_step_tolerance > FP{0.0}`. I'd suggest an assert here as well as documentation on the getter.

reneSchm (changes requested):
Nice addition, though I'd like some changes.

cpp/memilio/math/integrator.h
reneSchm: That is, for m_last_step_tolerance <= 0 we get an infinite loop, for m_last_step_tolerance >= 1, we simply don't do anything. So the valid range is (0, 1), though >=1 is less problematic.

